PR #678, opened 2026-04-15
feat(perf): features for evaluation

nokosaaan:
## Description
Since period information is required during evaluation, I have provided a feature to enable evaluation outside of the application.
`period-index-propagation` is normally turned off, which is consistent with the behavior of the traditional `awkernel`. To enable it, simply add `period-index-propagation` to the `default` in the `[features]` section of `kernel/Cargo.toml`.

I have added `perf` as a feature for `rv32` and `rv64`, but this is a temporary measure to pass tests on GitHub and is not intended to permanently add `perf` functionality to these targets. 
**The root cause of this issue has not yet been resolved and will be tracked in a separate issue.**

I have updated `pubsub.rs` to allow cycle propagation without altering existing trait boundaries, APIs, or the test application. Please refer to the file attached to the PR test section for the execution results.

## Related links
[sample code](https://github.com/tier4/awkernel/blob/dag_perf/applications/tests/test_dag/src/lib.rs)

## How was this PR tested?
[no_feature_compile.log](https://github.com/user-attachments/files/26745137/no_feature_compile.log)
[no_feature_execute.log](https://github.com/user-attachments/files/26745134/no_feature_execute.log)
[feature_compile.log](https://github.com/user-attachments/files/26745168/feature_compile.log)
[feature_execute.log](https://github.com/user-attachments/files/26745169/feature_execute.log)


awkernel_async_lib/src/task.rs
```diff
@@ -598,7 +1020,24 @@ pub mod perf {
...
+        if let Some(task_id) = task::get_current_task(cpu_id) {
+            if let Some(task) = task::get_task(task_id) {
+                let dag_info = task.info.lock(&mut task::MCSNode::new()).get_dag_info();
+                if dag_info.is_some() {
+                    // DAG task-specific handling
+                    update_time_and_state_for_dag(PerfState::Interrupt);
+                } else {
+                    // Normal task-specific handling
+                    update_time_and_state(PerfState::Interrupt);
+                }
```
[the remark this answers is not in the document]
nokosaaan: This means I didn't implement complete if-else handling.
I noticed that Copilot is correct. I'll fix it.

awkernel_async_lib/src/pubsub.rs
```diff
@@ -756,12 +785,25 @@ pub trait MultipleReceiver {
...
+
+    #[cfg(feature = "need-get-period")]
+    fn send_all_with_meta(
+        &self,
+        item: Self::Item,
+        pub_id: u32,
+        index: usize,
+        node_id: u32,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + '_>>;
 }
```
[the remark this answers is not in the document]
nokosaaan: This means I should delete `#[cfg(not(feature = "need-get-period"))]` above `send_all`.
I noticed that Copilot is correct. I'll fix it.

awkernel_async_lib/src/dag.rs
```diff
@@ -966,13 +994,35 @@ where
...
+                        release_time,
+                        dag_info.dag_id.clone(),
+                    );
+                }
+                let results = f();
+                publishers
+                    .send_all_with_meta(results, 0, index, dag_info.node_id)
+                    .await;
+                increment_period_count(dag_info.dag_id.clone() as usize);
+            }
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3101895482)

awkernel_async_lib/src/task.rs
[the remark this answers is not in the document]
nokosaaan: @copilot 
I fixed it.

awkernel_async_lib/src/pubsub.rs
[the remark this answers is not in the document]
nokosaaan: @copilot 
I fixed it.

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +497,343 @@ pub mod perf {
...
+    }
+
+    pub fn get_period_count(dag_id: usize) -> u32 {
+        assert!(dag_id < MAX_DAGS, "DAG ID out of bounds");
+        PERIOD_COUNT[dag_id].load(Ordering::Relaxed)
+    }
+
+    pub fn update_pre_send_outer_timestamp_at(index: usize, new_timestamp: u64, dag_id: u32) {
+        assert!(index < MAX_LOGS, "Timestamp index out of bounds");
+
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3101895492)

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +497,343 @@ pub mod perf {
...
+            "Latency",
+            "Absolute Deadline",
+            "Relative Deadline"
+        );
+
+        log::info!("-----|--------|----------------|----------------|----------------|--------------------|--------------------|--------------------|--------------------");
+
+        for i in 0..MAX_LOGS {
+            for j in 1..MAX_DAGS {
+                let pre_send_outer = match &*send_outer_opt {
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3101895534)

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +497,343 @@ pub mod perf {
...
+
+        let mut node = MCSNode::new();
+        let mut recorder_opt = SEND_OUTER_TIMESTAMP.lock(&mut node);
+
+        let recorder = recorder_opt.get_or_insert_with(|| Box::new([[0; MAX_DAGS]; MAX_LOGS]));
+
+        if recorder[index][dag_id as usize] == 0 {
+            recorder[index][dag_id as usize] = new_timestamp;
+        }
+    }
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3101895600)

awkernel_async_lib/src/time_interval.rs
```diff
@@ -95,13 +97,23 @@ pub enum MissedTickBehavior {
...
     assert!(!period.is_zero(), "`period` must be non-zero.");
+    #[cfg(feature = "need-get-period")]
+    {
+        let index = get_period_count(_dag_id.clone() as usize) as usize;
+        // [start] cycle deviation index == 0 (basis of cycle deviation)
+        update_pre_send_outer_timestamp_at(
+            index,
+            start.uptime().as_nanos() as u64,
+            _dag_id.clone(),
+        );
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3107204674)

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        let mut node = MCSNode::new();
+        let mut recorder_opt = SEND_OUTER_TIMESTAMP.lock(&mut node);
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, new_timestamp);
+        }
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
This change could significantly alter how logs are displayed, so I’d prefer to keep them as they are if possible.

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        let mut node = MCSNode::new();
+        let mut recorder_opt = ABSOLUTE_DEADLINE.lock(&mut node);
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, deadline);
+        }
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
This change could significantly alter how logs are displayed, so I’d prefer to keep them as they are if possible.

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+        let mut node = MCSNode::new();
+        let mut recorder_opt = SUBSCRIBE.lock(&mut node);
+
+        let recorder = recorder_opt.get_or_insert_with(|| Box::new(PubSubTable::default()));
+        let sub_id = sub_id as usize;
+
+        if recorder.timestamps[index][sub_id][node_id_usize] == 0 {
+            recorder.timestamps[index][sub_id][node_id_usize] = new_timestamp;
+            recorder.mark_used(index, sub_id, node_id_usize);
+        }
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
Same wrap-around issue as publish_timestamp_at: once timestamps[index][sub_id][node_id] is non-zero, later cycles that reuse the same normalized index won’t be recorded. If you intend a ring buffer, you likely want overwrite semantics (or explicit clearing on wrap).

awkernel_async_lib/src/time_interval.rs
```diff
@@ -95,13 +97,23 @@ pub enum MissedTickBehavior {
 /// }
 /// ```
 ///
-pub fn interval(period: Duration) -> Interval {
+pub fn interval(period: Duration, dag_id: u32) -> Interval {
     assert!(!period.is_zero(), "`period` must be non-zero.");
-    interval_at(Time::now(), period)
+    interval_at(Time::now(), period, dag_id)
 }
```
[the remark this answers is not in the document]
nokosaaan: @copilot apply changes based on [this feedback](https://github.com/tier4/awkernel/pull/678#discussion_r3107204692)

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,95 @@ where
...
+
+    #[test]
+    fn send_with_meta_propagates_period_to_receiver() {
+        block_on(async {
+            let (publisher, subscriber) = create_pubsub::<u32>(Attribute::default());
+            publisher.send_with_meta(42, 1, 7, 99).await;
+
+            let (value, period) = subscriber.recv_with_period().await;
+            assert_eq!(value, 42);
+            assert_eq!(period, 7);
```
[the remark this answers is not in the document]
nokosaaan: I fixed it on the latest commit.

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, deadline);
+        }
+    }
+
+    pub fn publish_timestamp_at(index: usize, new_timestamp: u64, pub_id: u32, node_id: u32) {
+        let index = normalize_log_index(index);
+        assert!((pub_id as usize) < MAX_PUBSUB, "Publish ID out of bounds");
+
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
Since the pub_id is essentially fixed, placing an assert here is simply a matter of convention. Unless the user intentionally modifies it, this will not typically result in an error.

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+        let mut node = MCSNode::new();
+        let mut recorder_opt = PUBLISH.lock(&mut node);
+
+        let recorder = recorder_opt.get_or_insert_with(|| Box::new(PubSubTable::default()));
+        let pub_id = pub_id as usize;
+
+        if recorder.timestamps[index][pub_id][node_id_usize] == 0 {
+            recorder.timestamps[index][pub_id][node_id_usize] = new_timestamp;
+            recorder.mark_used(index, pub_id, node_id_usize);
+        }
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
Since changing the MAX_LOGS value is within the range where we can manually prevent a panic from occurring, it is not practical when considering the cost of making that change. I will fix this if it is pointed out in other reviews.

kernel/src/main.rs
```diff
@@ -74,7 +74,17 @@ fn main<Info: Debug>(kernel_info: KernelInfo<Info>) {
...
         unsafe { awkernel_lib::cpu::init_sleep() };
 
+        #[cfg(feature = "need-get-period")]
+        let mut last_print = awkernel_lib::time::Time::now();
+
         loop {
+            #[cfg(feature = "need-get-period")]
+            if last_print.elapsed().as_secs() >= 30 {
+                awkernel_async_lib::task::perf::print_timestamp_table();
+                // awkernel_async_lib::task::perf::print_pubsub_table();
```
kobayu858: Please delete this if it is unnecessary. If necessary, please leave a comment.

awkernel_async_lib/Cargo.toml
```diff
@@ -45,3 +45,4 @@ no_preempt = []
 spinlock = ["awkernel_lib/spinlock"]
 clippy = []
 perf = []
+need-get-period = ["perf"]
```
kobayu858: While the name `need-get-period` suggests that we `need to retrieve the duration`, what is actually being sent and received is not the duration but the index.
Some more accurate name candidates might include:

- period-index-propagation
- cycle-tracing
- dag-latency-measure

I believe these would better reflect the actual functionality.

kernel/Cargo.toml
```diff
@@ -92,12 +93,14 @@ x86 = [
...
     "no_std_unwinding",
     "awkernel_lib/rv32",
     "dep:ns16550a",
     "awkernel_drivers/rv32",
 ]
 rv64 = [
+    "perf",
     "no_std_unwinding",
     "awkernel_lib/rv64",
     "dep:ns16550a",
```
kobayu858, line 110: This change enables `perf` measurements for RISC-V targets, but it doesn't seem directly related to `need-get-period`. Was this intentional, or did it just get included incidentally?

kernel/Cargo.toml
```diff
@@ -92,12 +93,14 @@ x86 = [
     "awkernel_drivers/uart_16550",
 ]
 rv32 = [
+    "perf",
```
kobayu858, line 99: Is this a change intended to enable perf measurements on RISC-V targets, or is it unrelated to `need-get-period` and was simply added as a side note?

awkernel_async_lib/src/time_interval.rs
```diff
@@ -95,13 +97,19 @@ pub enum MissedTickBehavior {
 /// }
 /// ```
 ///
-pub fn interval(period: Duration) -> Interval {
+pub fn interval(period: Duration, dag_id: u32) -> Interval {
```
kobayu858: The argument name for `interval()` is not consistent with `dag_id` (without the underscore).

awkernel_async_lib/src/dag.rs
```diff
@@ -926,10 +933,31 @@ where
             Args::create_subscribers(subscribe_topic_names, Attribute::default());
 
         loop {
-            let args: <<Args as VectorToSubscribers>::Subscribers as MultipleReceiver>::Item =
-                subscribers.recv_all().await;
-            let results = f(args);
-            publishers.send_all(results).await;
+            #[cfg(feature = "need-get-period")]
+            {
+                let (args, count_st): (
```
kobayu858: The naming of `count_st` and related variables/functions is inconsistent throughout the codebase. The suffix st is unclear and its meaning cannot be inferred from context. Additionally, period, index, count, and st are used interchangeably to represent the same concept — "which cycle number this is" — making the code difficult to read and maintain.

`count_st` → `period_index`

awkernel_async_lib/src/pubsub.rs
```diff
@@ -276,8 +283,16 @@ impl<'a, T: Send> Sender<'a, T> {
             subscribers: Default::default(),
             state: SenderState::Start,
             timestamp: awkernel_lib::time::Time::now(),
+            #[cfg(feature = "need-get-period")]
+            index: 0,
         }
     }
+
+    #[cfg(feature = "need-get-period")]
+    pub(super) fn with_period(mut self, index: u32) -> Self {
```
kobayu858: `with_period` → `with_period_index`

awkernel_async_lib/src/pubsub.rs
```diff
@@ -276,8 +283,16 @@ impl<'a, T: Send> Sender<'a, T> {
             subscribers: Default::default(),
             state: SenderState::Start,
             timestamp: awkernel_lib::time::Time::now(),
+            #[cfg(feature = "need-get-period")]
+            index: 0,
```
kobayu858: `index`→`period_index`

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
         sender.await;
         r#yield().await;
     }
+
+    #[cfg(feature = "need-get-period")]
+    pub async fn send_with_meta(&self, data: T, pub_id: u32, index: usize, node_id: u32) {
```
kobayu858: `send_all_with_meta` → `send_all_with_period_index` (meta is vague)

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+        // [start] pubsub communication latency
+        let start = awkernel_lib::time::Time::now().uptime().as_nanos() as u64;
+        publish_timestamp_at(index, start, pub_id, node_id);
+        let period = match u32::try_from(index) {
```
kobayu858: `period`→`period_index`

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+            Ok(period) => period,
+            Err(_) => {
+                log::warn!(
+                    "Period index {} exceeds u32::MAX; saturating period metadata",
```
kobayu858: "saturating period metadata" → "saturating period index"

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+            let mut context = Context::from_waker(&waker);
+            match Pin::as_mut(&mut future).poll(&mut context) {
+                Poll::Ready(output) => return output,
+                Poll::Pending => {}
+            }
+        }
+    }
+
+    #[test]
+    fn send_with_meta_propagates_period_to_receiver() {
```
kobayu858: send_with_period_index_propagates_period_index_to_receiver

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+            publisher.send_with_meta(42, 1, 7, 99).await;
+
+            let received = subscriber.recv().await;
+            assert_eq!(received.data, 42);
+            assert_eq!(received.index, 7);
+        });
+    }
+
+    #[test]
+    fn tuple_recv_all_with_period_returns_shared_period() {
```
kobayu858: tuple_recv_all_with_period_index_returns_shared_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+        block_on(async {
+            let (publisher1, subscriber1) = create_pubsub::<u32>(Attribute::default());
+            let (publisher2, subscriber2) = create_pubsub::<u32>(Attribute::default());
+
+            publisher1.send_with_meta(10, 11, 3, 21).await;
+            publisher2.send_with_meta(20, 12, 3, 22).await;
+
+            let ((value1, value2), period) =
```
kobayu858: period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,94 @@ where
...
+
+    #[test]
+    fn send_with_meta_propagates_period_to_receiver() {
+        block_on(async {
+            let (publisher, subscriber) = create_pubsub::<u32>(Attribute::default());
+            publisher.send_with_meta(42, 1, 7, 99).await;
+
+            let received = subscriber.recv().await;
+            assert_eq!(received.data, 42);
+            assert_eq!(received.index, 7);
```
kobayu858: received.period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -756,12 +861,24 @@ pub trait MultipleReceiver {
     type Item;
 
     fn recv_all(&self) -> Pin<Box<dyn Future<Output = Self::Item> + Send + '_>>;
+
+    #[cfg(feature = "need-get-period")]
+    fn recv_all_with_period(&self) -> Pin<Box<dyn Future<Output = (Self::Item, u32)> + Send + '_>>;
```
kobayu858: recv_all_with_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -756,12 +861,24 @@ pub trait MultipleReceiver {
...
 }
 
 pub trait MultipleSender {
     type Item;
 
     fn send_all(&self, item: Self::Item) -> Pin<Box<dyn Future<Output = ()> + Send + '_>>;
+
+    #[cfg(feature = "need-get-period")]
+    fn send_all_with_meta(
```
kobayu858: send_all_with_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -834,6 +951,13 @@ macro_rules! impl_async_receiver_for_tuple {
             fn recv_all(&self) -> Pin<Box<dyn Future<Output = Self::Item> + Send + '_>> {
                 Box::pin(async move{})
             }
+
+            #[cfg(feature = "need-get-period")]
+            fn recv_all_with_period(
```
kobayu858: recv_all_with_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -842,6 +966,17 @@ macro_rules! impl_async_receiver_for_tuple {
             fn send_all(&self, _item: Self::Item) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
                 Box::pin(async move{})
             }
+
+            #[cfg(feature = "need-get-period")]
+            fn send_all_with_meta(
```
kobayu858: send_all_with_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -854,6 +989,34 @@ macro_rules! impl_async_receiver_for_tuple {
                     ($($idx.recv().await.data,)+)
                 })
             }
+
+            #[cfg(feature = "need-get-period")]
+            fn recv_all_with_period(
```
kobayu858: recv_all_with_period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -854,6 +989,34 @@ macro_rules! impl_async_receiver_for_tuple {
...
+                &self,
+            ) -> Pin<Box<dyn Future<Output = (Self::Item, u32)> + Send + '_>> {
+                let ($($idx,)+) = self;
+                Box::pin(async move {
+                    let mut period: Option<u32> = None;
+                    $(
+                        let item = $idx.recv().await;
+                        match period {
```
kobayu858: period_index

awkernel_async_lib/src/pubsub.rs
```diff
@@ -868,6 +1031,23 @@ macro_rules! impl_async_receiver_for_tuple {
                     )+
                 })
             }
+
+            #[cfg(feature = "need-get-period")]
+            fn send_all_with_meta(
```
kobayu858: send_all_with_period_index

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+    pub fn get_period_count(dag_id: u32) -> u32 {
+        let mut node = MCSNode::new();
+        let period_count = PERIOD_COUNT.lock(&mut node);
+        period_count
+            .get(&dag_id)
+            .map(|count| count.load(core::sync::atomic::Ordering::Relaxed))
+            .unwrap_or(0)
+    }
+
+    pub fn increment_period_count(dag_id: u32) -> u32 {
```
kobayu858: increment_period_index

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+    type DagTimestampTable = [DagTimestampMap; MAX_LOGS];
+
+    static SEND_OUTER_TIMESTAMP: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static RECV_OUTER_TIMESTAMP: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static ABSOLUTE_DEADLINE: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static RELATIVE_DEADLINE: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+
+    pub static PERIOD_COUNT: Mutex<BTreeMap<u32, AtomicU32>> = Mutex::new(BTreeMap::new());
+
+    pub fn get_period_count(dag_id: u32) -> u32 {
```
kobayu858: get_period_index

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+    type DagTimestampMap = BTreeMap<u32, u64>;
+    type DagTimestampTable = [DagTimestampMap; MAX_LOGS];
+
+    static SEND_OUTER_TIMESTAMP: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static RECV_OUTER_TIMESTAMP: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static ABSOLUTE_DEADLINE: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+    static RELATIVE_DEADLINE: Mutex<Option<Box<DagTimestampTable>>> = Mutex::new(None);
+
+    pub static PERIOD_COUNT: Mutex<BTreeMap<u32, AtomicU32>> = Mutex::new(BTreeMap::new());
```
kobayu858: PERIOD_INDEX

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+    static PUBLISH: Mutex<Option<Box<PubSubTable>>> = Mutex::new(None);
+    static SUBSCRIBE: Mutex<Option<Box<PubSubTable>>> = Mutex::new(None);
+
+    #[inline(always)]
+    fn normalize_log_index(index: usize) -> usize {
+        index % MAX_LOGS
+    }
+
+    pub fn update_pre_send_outer_timestamp_at(index: usize, new_timestamp: u64, dag_id: u32) {
```
kobayu858: update_cycle_start_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, new_timestamp);
+        }
+    }
+
+    pub fn update_fin_recv_outer_timestamp_at(index: usize, new_timestamp: u64, dag_id: u32) {
```
kobayu858: update_cycle_end_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+        let recorder = recorder_opt.get_or_insert_with(|| Box::new(PubSubTable::default()));
+        let pub_id = pub_id as usize;
+
+        if recorder.timestamps[index][pub_id][node_id_usize] == 0 {
+            recorder.timestamps[index][pub_id][node_id_usize] = new_timestamp;
+            recorder.mark_used(index, pub_id, node_id_usize);
+        }
+    }
+
+    pub fn subscribe_timestamp_at(index: usize, new_timestamp: u64, sub_id: u32, node_id: u32) {
```
kobayu858: record_subscribe_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, deadline);
+        }
+    }
+
+    pub fn publish_timestamp_at(index: usize, new_timestamp: u64, pub_id: u32, node_id: u32) {
```
kobayu858: record_publish_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[index].get(&dag_id).copied().unwrap_or(0) == 0 {
+            recorder[index].insert(dag_id, deadline);
+        }
+    }
+
+    pub fn update_relative_deadline_timestamp_at(index: usize, deadline: u64, dag_id: u32) {
```
kobayu858: update_relative_deadline_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+        let mut node = MCSNode::new();
+        let mut recorder_opt = RECV_OUTER_TIMESTAMP.lock(&mut node);
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        recorder[index].insert(dag_id, new_timestamp);
+    }
+
+    pub fn update_absolute_deadline_timestamp_at(index: usize, deadline: u64, dag_id: u32) {
```
kobayu858: update_absolute_deadline_timestamp

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+        }
+    }
+
+    static PUBLISH: Mutex<Option<Box<PubSubTable>>> = Mutex::new(None);
+    static SUBSCRIBE: Mutex<Option<Box<PubSubTable>>> = Mutex::new(None);
+
+    #[inline(always)]
+    fn normalize_log_index(index: usize) -> usize {
+        index % MAX_LOGS
+    }
```
kobayu858: fn to_ring_buffer_index(period_index: usize) -> usize {
    period_index % MAX_LOGS
}

awkernel_async_lib/src/task.rs
```diff
@@ -490,23 +496,405 @@ pub mod perf {
...
+
+    pub fn update_pre_send_outer_timestamp_at(index: usize, new_timestamp: u64, dag_id: u32) {
+        let index = normalize_log_index(index);
```
kobayu858: let log_index = to_ring_buffer_index(period_index);

awkernel_async_lib/src/task.rs
```diff
@@ -554,6 +942,101 @@ pub mod perf {
...
+                PerfState::Task => unsafe {
+                    let t = read_volatile(&TASK_TIME[cpu_id]);
+                    write_volatile(&mut TASK_TIME[cpu_id], t + diff);
+                    let c = read_volatile(&TASK_COUNT[cpu_id]);
+                    write_volatile(&mut TASK_COUNT[cpu_id], c + 1);
+                    let wcet = read_volatile(&TASK_WCET[cpu_id]);
+                    write_volatile(&mut TASK_WCET[cpu_id], wcet.max(diff));
+                },
+                PerfState::Interrupt => unsafe {
+                    // log::info!("PreemptionTime CPU:{} Diff:{}", cpu_id, diff);
```
kobayu858: remove

awkernel_async_lib/src/task.rs
```diff
@@ -554,6 +942,101 @@ pub mod perf {
...
+                    let t = read_volatile(&INTERRUPT_TIME[cpu_id]);
+                    write_volatile(&mut INTERRUPT_TIME[cpu_id], t + diff);
+                    let c = read_volatile(&INTERRUPT_COUNT[cpu_id]);
+                    write_volatile(&mut INTERRUPT_COUNT[cpu_id], c + 1);
+                    let wcet = read_volatile(&INTERRUPT_WCET[cpu_id]);
+                    write_volatile(&mut INTERRUPT_WCET[cpu_id], wcet.max(diff));
+                },
+                PerfState::ContextSwitch => unsafe {
+                    // log::info!("ContextSwitchTime CPU:{} Diff:{}", cpu_id, diff);
```
kobayu858: remove

awkernel_async_lib/src/task.rs
```diff
@@ -554,6 +942,101 @@ pub mod perf {
...
+                    let t = read_volatile(&CONTEXT_SWITCH_TIME[cpu_id]);
+                    write_volatile(&mut CONTEXT_SWITCH_TIME[cpu_id], t + diff);
+                    let c = read_volatile(&CONTEXT_SWITCH_COUNT[cpu_id]);
+                    write_volatile(&mut CONTEXT_SWITCH_COUNT[cpu_id], c + 1);
+                    let wcet = read_volatile(&CONTEXT_SWITCH_WCET[cpu_id]);
+                    write_volatile(&mut CONTEXT_SWITCH_WCET[cpu_id], wcet.max(diff));
+                },
+                PerfState::ContextSwitchMain => unsafe {
+                    // log::info!("ContextSwitchMainTime CPU:{} Diff:{}", cpu_id, diff);
```
kobayu858: remove

kobayu858 (changes requested):

kernel/Cargo.toml
nokosaaan: This change was added to pass tests on the RISC-V target’s GitHub repository; it was not intended to enable performance measurement. Ideally, we would not add it if the tests could be passed without it, but since we have not yet identified a way to pass the tests without it, the code remains in its current state.

awkernel_async_lib/src/time_interval.rs
nokosaaan: Is the point of this comment that the function name isn’t appropriate because it includes `dag_id`? Or is it suggesting that we should align it with `interval_at`, which includes an underscore in `dag_id`?

kernel/src/main.rs
nokosaaan: I deleted it.

kernel/Cargo.toml
nokosaaan: This change was added to pass tests on the rv32 and rv64 target’s GitHub repository; it was not intended to enable performance measurement. Ideally, we would not add it if the tests could be passed without it, but since we have not yet identified a way to pass the tests without it, the code remains in its current state.

awkernel_async_lib/src/dag.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/Cargo.toml
nokosaaan: I renamed feature `need-get-period` to `period-index-propagation`

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I removed it.

awkernel_async_lib/src/task.rs
nokosaaan: I removed it.

awkernel_async_lib/src/task.rs
nokosaaan: I removed it.

awkernel_async_lib/src/time_interval.rs
kobayu858: Unless there is a specific reason not to, I would like to standardize the argument names.

interval_at(start: Time, period: Duration, _dag_id: u32)
pub fn interval(period: Duration, dag_id: u32) 

In other words, either `_dag_id` or `dag_id` is fine, so please standardize on one of them.

awkernel_async_lib/Cargo.toml
kobayu858: There are still many instances where “need-get-period” remains, so please check them.

kobayu858: It seems I was looking at an old version history. I'm retracting the above.

awkernel_async_lib/src/dag.rs
```diff
@@ -67,6 +67,13 @@ use crate::{
     time_interval::interval,
     Attribute, MultipleReceiver, MultipleSender, VectorToPublishers, VectorToSubscribers,
 };
+
+#[cfg(feature = "period-index-propagation")]
+use crate::task::perf::{
+    get_period_count, increment_period_count, record_subscribe_timestamp,
```
kobayu858: `dag.rs` calls `get_period_count` and `increment_period_count`, but `task.rs` only defines `get_period_index` and `increment_period_index`. These need to match.

awkernel_async_lib/src/dag.rs
```diff
@@ -926,10 +933,31 @@ where
...
-            let args: <<Args as VectorToSubscribers>::Subscribers as MultipleReceiver>::Item =
-                subscribers.recv_all().await;
-            let results = f(args);
-            publishers.send_all(results).await;
+            #[cfg(feature = "period-index-propagation")]
+            {
+                let (args, period_index): (
+                    <<Args as VectorToSubscribers>::Subscribers as MultipleReceiver>::Item,
+                    u32,
+                ) = subscribers.recv_all_with_period().await;
```
kobayu858: `dag.rs` calls `subscribers.recv_all_with_period()`, but the `MultipleReceiver` trait in `pubsub.rs` defines the method as `recv_all_with_period_index`. This will also fail to compile.

awkernel_async_lib/src/time_interval.rs
nokosaaan: I have standardized it to`_dag_id`.

awkernel_async_lib/src/dag.rs
nokosaaan: I changed the function name from an unused one to the one currently in use.

awkernel_async_lib/src/dag.rs
nokosaaan: I changed the function name from an unused one to the one currently in use.

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,95 @@ where
...
+            }
+        };
+        let sender = Sender::new(self, data).with_period_index(period_index);
+        sender.await;
+        r#yield().await;
+    }
 }
 
+#[cfg(all(test, feature = "period-index-propagation"))]
+mod need_get_period_tests {
```
kobayu858: If we want to match the feature name, we should rename it to something like period_index_propagation_tests.

awkernel_async_lib/src/time_interval.rs
```diff
@@ -95,13 +97,19 @@ pub enum MissedTickBehavior {
 /// }
 /// ```
 ///
-pub fn interval(period: Duration) -> Interval {
+pub fn interval(period: Duration, _dag_id: u32) -> Interval {
```
kobayu858: **Breaking change in time_interval::interval API**
Please check whether `interval` and` interval_at` are called directly from outside this crate (e.g. from the kernel side or other crates). If they are only called internally from　`awkernel_async_lib/src/dag.rs`, restrict their visibility to `pub(crate)`.

awkernel_async_lib/src/pubsub.rs
```diff
@@ -386,8 +405,95 @@ where
...
+        }
+
+        let waker = unsafe { Waker::from_raw(raw_waker()) };
+        let mut future = Box::pin(future);
+
+        loop {
+            let mut context = Context::from_waker(&waker);
+            match Pin::as_mut(&mut future).poll(&mut context) {
+                Poll::Ready(output) => return output,
+                Poll::Pending => {}
```
kobayu858, line 464: wake is a no-op, so returning `Poll::Pending` would cause a busy-loop. However, since this test follows a single-threaded, send-before-receive ordering where the send always completes before the recv, it should not actually hang in practice. Please add a comment explicitly stating this test's assumptions (single-threaded, send-then-receive ordering).

awkernel_async_lib/src/pubsub.rs
```diff
@@ -854,6 +992,34 @@ macro_rules! impl_async_receiver_for_tuple {
...
+                        let item = $idx.recv().await;
+                        match period_index {
+                            Some(expected) => {
+                                assert!(
+                                    expected == item.period_index,
+                                    "recv_all_with_period_index received mismatched periods: expected {}, got {}",
+                                    expected,
+                                    item.period_index
+                                );
+                            }
```
kobayu858: Receiving messages with different period_index values from multiple upstream nodes will cause a panic.This can happen in the following scenarios:
- Upstream nodes start at different times
- One upstream is delayed and drops a message
- A DAG branch-and-merge has paths of different lengths

However, crashing the entire system is too heavy-handed. Consider using `log::warn!` to skip mismatched messages, or returning a `Result` instead. At a minimum, the conditions under which this assert panics and the rationale behind it should be documented in a doc comment.

awkernel_async_lib/src/task.rs
```diff
@@ -584,6 +982,98 @@ pub mod perf {
...
+            let wcet = read_volatile(&PERF_WCET[cpu_id]);
+            write_volatile(&mut PERF_WCET[cpu_id], wcet.max(cnt - end));
+
+            // State transition.
+            write_volatile(&mut START_TIME[cpu_id], cnt);
+            write_volatile(&mut PERF_STATES[cpu_id], next_state as u8);
+        }
+    }
+
+    fn update_time_and_state_for_dag(next_state: PerfState) {
```
kobayu858: `update_time_and_state_for_dag` is an exact duplicate of `update_time_and_state` — not a single line differs. Despite the name implying DAG-specific behavior, both paths in `start_interrupt` and `start_context_switch` (branched via `dag_info.is_some()`) ultimately reach the same logic.

Please remove `update_time_and_state_for_dag` and consolidate both paths to call `update_time_and_state` directly.

awkernel_async_lib/src/task.rs
```diff
@@ -628,7 +1118,27 @@ pub mod perf {
     pub fn start_interrupt() -> PerfState {
         let cpu_id = awkernel_lib::cpu::cpu_id();
         let previous: PerfState = unsafe { read_volatile(&PERF_STATES[cpu_id]) }.into();
-        update_time_and_state(PerfState::Interrupt);
+
+        // Check the current task
+        if let Some(task_id) = task::get_current_task(cpu_id) {
+            if let Some(task) = task::get_task(task_id) {
+                let dag_info = task.info.lock(&mut task::MCSNode::new()).get_dag_info();
```
kobayu858: In start_interrupt and start_context_switch, an MCS lock is acquired via task.info.lock(). Taking a lock on this high-frequency interrupt handler path is problematic both in terms of deadlock risk and performance overhead.
That said, as noted in a separate comment, removing update_time_and_state_for_dag and consolidating into update_time_and_state would eliminate the dag_info.is_some() branch entirely and with it, this lock acquisition as well.

awkernel_async_lib/src/task.rs
```diff
@@ -520,23 +526,415 @@ pub mod perf {
...
+            .entry(dag_id)
+            .or_insert_with(|| AtomicU32::new(0));
+        count.fetch_add(1, core::sync::atomic::Ordering::Relaxed) + 1
+    }
+    //pubsub
+    const MAX_PUBSUB: usize = 3;
+    const MAX_NODES: usize = 20;
+    #[derive(Clone)]
+    struct PubSubTable {
+        timestamps: [[[u64; MAX_NODES]; MAX_PUBSUB]; MAX_LOGS],
```
kobayu858, line 612: Please add comments explaining the rationale behind `MAX_LOGS`, `MAX_PUBSUB`, and `MAX_NODES`, and whether there is any room for tuning them. As it stands, these constants result in a static allocation of roughly `3.75 MB (8192 × 3 × 20 × 8 bytes)`, which is fairly large and warrants some justification.

kernel/Cargo.toml
kobayu858: Please address the following:
① Update the PR description — State explicitly that perf was added as a temporary workaround because tests do not pass on rv32/rv64 without it, and that the root cause has not yet been investigated. Leaving the intent undocumented will confuse anyone who looks at this later.
② Add a TODO comment
```toml
rv32 = [
    # TODO: Remove once a way to pass tests without `perf` is found
    "perf",
    ...
]
```
③ Open a separate issue — It is recommended to track the investigation of why tests fail without perf on RISC-V in a dedicated issue. Left as-is, RISC-V builds will permanently carry the perf feature, which is likely unintended.

awkernel_async_lib/src/task.rs
```diff
@@ -520,23 +526,415 @@ pub mod perf {
...
+            "Index",
+            "DAG-ID",
+            "Send-Outer",
+            "Recv-Outer",
+            "Latency",
+            "Absolute Deadline",
+            "Relative Deadline"
+        );
+
+        log::info!("-----|--------|----------------|----------------|----------------|--------------------|--------------------|--------------------|--------------------");
```
kobayu858: There are 9 separator columns but only 7 columns in the header and data rows. Please verify that the log output is visually aligned, and if it is misaligned, fix the separator count to match. If the discrepancy is intentional, please add a comment explaining why.

awkernel_async_lib/src/task.rs
```diff
@@ -520,23 +526,415 @@ pub mod perf {
...
+    }
+
+    pub fn record_publish_timestamp(
+        period_index: usize,
+        new_timestamp: u64,
+        pub_id: u32,
+        node_id: u32,
+    ) {
+        let log_index = to_ring_buffer_index(period_index);
+        assert!((pub_id as usize) < MAX_PUBSUB, "Publish ID out of bounds");
```
kobayu858: Within the same function, out-of-bounds checks for node_id are handled gracefully via log::warn! + return, while pub_id/sub_id use assert! and panic — an inconsistent approach. Please unify all three to use log::warn! + return.

awkernel_async_lib/src/task.rs
```diff
@@ -520,23 +526,415 @@ pub mod perf {
...
+    }
+
+    pub fn record_subscribe_timestamp(
+        period_index: usize,
+        new_timestamp: u64,
+        sub_id: u32,
+        node_id: u32,
+    ) {
+        let log_index = to_ring_buffer_index(period_index);
+        assert!((sub_id as usize) < MAX_PUBSUB, "Subscribe ID out of bounds");
```
kobayu858: Within the same function, out-of-bounds checks for node_id are handled gracefully via log::warn! + return, while pub_id/sub_id use assert! and panic — an inconsistent approach. Please unify all three to use log::warn! + return.

kobayu858 (changes requested):

kernel/Cargo.toml
nokosaaan: I addressed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/time_interval.rs
nokosaaan: After confirming that it is not used outside of `dag.rs`, I restricted the function's visibility to `pub(crate)`.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I added a comment.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I fixed it (assert → log::warn!).

awkernel_async_lib/src/task.rs
nokosaaan: If we remove `update_time_and_state_for_dag` and merge it into `update_time_and_state`, do we no longer need to worry about the possibility that the code in `start_interrupt`, `start_context_switch`, and `start_context_switch_main` might be perceived as containing redundant if-else branches?
Or is the intent of this comment to eliminate all if-else branches and simply change the code to call `update_time_and_state`?

awkernel_async_lib/src/task.rs
nokosaaan: For now, I’ve removed the if-else handling and made it so that `update_time_and_state` is called regardless of which task is received. Please let me know if this change was unintended.

awkernel_async_lib/src/task.rs
nokosaaan: For now, I’ve removed the if-else handling and made it so that `update_time_and_state` is called regardless of which task is received. Please let me know if this change was unintended.

awkernel_async_lib/src/task.rs
nokosaaan: `MAX_LOGS` and `MAX_NODES` are adjustable constants, and I have left a comment to that effect. (Since `MAX_LOGS` is usually fine even if it is a bit smaller, I changed the default to `2048`. The value of `2048` was chosen based on [this branch](https://github.com/tier4/awkernel/blob/36a5fb20f3277c220eac05240d79c9d17af09bbe/awkernel_async_lib/src/task.rs#L512).)

nokosaaan: The value of `MAX_NODES` was changed to `5`.
This is because the `test_dag` test application, which runs the DAG, has `5 nodes`. (If the user has configured a DAG with `10 nodes`, this value can be changed to `10`, or if the maximum number of nodes is updated frequently, it can be raised to a value that poses no issues.)

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/task.rs
kobayu858: Yes, that's exactly what was intended. Since update_time_and_state_for_dag and update_time_and_state were identical, the if-else branches in start_interrupt, start_context_switch, and start_context_switch_main had no practical effect. Removing the branches and calling update_time_and_state directly is the correct fix. Thank you for making the change.

If you have a specific reason for wanting to keep `update_time_and_state_for_dag`, please let me know. In that case, please revert the change.

kernel/Cargo.toml
kobayu858: Please add a link to the corresponding issue in the TODO comment.

awkernel_async_lib/src/task.rs
```diff
@@ -520,23 +525,433 @@ pub mod perf {
...
+    pub fn update_cycle_start_timestamp(period_index: usize, new_timestamp: u64, dag_id: u32) {
+        let log_index = to_ring_buffer_index(period_index);
+
+        let mut node = MCSNode::new();
+        let mut recorder_opt = SEND_OUTER_TIMESTAMP.lock(&mut node);
+
+        let recorder =
+            recorder_opt.get_or_insert_with(|| Box::new(core::array::from_fn(|_| BTreeMap::new())));
+
+        if recorder[log_index].get(&dag_id).copied().unwrap_or(0) == 0 {
```
kobayu858: Timestamp 0 is used as a sentinel value meaning "not recorded", but this cannot be distinguished from a valid value if the system uptime happens to be near 0 ns at startup (unlikely in practice, but theoretically possible). Please either consider an alternative approach or add a comment acknowledging this limitation.

awkernel_async_lib/src/pubsub.rs
```diff
@@ -854,6 +995,35 @@ macro_rules! impl_async_receiver_for_tuple {
...
+                &self,
+            ) -> Pin<Box<dyn Future<Output = (Self::Item, u32)> + Send + '_>> {
+                let ($($idx,)+) = self;
+                Box::pin(async move {
+                    let mut period_index: Option<u32> = None;
+                    $(
+                        let item = $idx.recv().await;
+                        match period_index {
+                            Some(expected) => {
+                                if expected != item.period_index {
```
kobayu858, line 1015: The mismatch handling here (adopt first, warn for rest) looks fine, but could we add a comment explaining the intent?

kernel/src/main.rs
```diff
@@ -74,7 +74,16 @@ fn main<Info: Debug>(kernel_info: KernelInfo<Info>) {
         // Enable awkernel_lib::cpu::sleep_cpu() and awkernel_lib::cpu::wakeup_cpu().
         unsafe { awkernel_lib::cpu::init_sleep() };
 
+        #[cfg(feature = "period-index-propagation")]
+        let mut last_print = awkernel_lib::time::Time::now();
+
         loop {
+            #[cfg(feature = "period-index-propagation")]
+            if last_print.elapsed().as_secs() >= 30 {
```
kobayu858: This is a magic number. Please either introduce a named constant such as const PERF_PRINT_INTERVAL_SECS: u64 = 30; or make it configurable.

awkernel_async_lib/src/task.rs
kobayu858: Please add a comment explaining the rationale behind MAX_LOGS = 2048 and clarifying that this is a ring buffer.

kobayu858 (changes requested):

awkernel_async_lib/src/task.rs
nokosaaan: I see. If calling the same function—whether in a regular task or a DAG task—doesn’t affect previous evaluation results, then I’m in favor of this change as well.

awkernel_async_lib/src/task.rs
nokosaaan: I added a comment.

awkernel_async_lib/src/task.rs
nokosaaan: I fixed it.

awkernel_async_lib/src/pubsub.rs
nokosaaan: I added a comment.

kernel/src/main.rs
nokosaaan: I fixed it.

kernel/Cargo.toml
nokosaaan: I added a link.

awkernel_async_lib/src/task.rs
nokosaaan: Specifically, instead of using a timestamp, we now determine whether a record exists based on whether the index key is present.

awkernel_async_lib/src/task.rs
kobayu858: memo:

```
    fn update_time_and_state(next_state: PerfState) {
        let end = awkernel_lib::delay::cpu_counter();
        let cpu_id = awkernel_lib::cpu::cpu_id();

        let state: PerfState = unsafe { read_volatile(&PERF_STATES[cpu_id]) }.into();
        if state == next_state {
            return;
        }

        let start = unsafe { read_volatile(&START_TIME[cpu_id]) };

        if start > 0 && start <= end {
            let diff = end - start;

            match state {
                PerfState::Kernel => unsafe {
                    let t = read_volatile(&KERNEL_TIME[cpu_id]);
                    write_volatile(&mut KERNEL_TIME[cpu_id], t + diff);
                    let c = read_volatile(&KERNEL_COUNT[cpu_id]);
                    write_volatile(&mut KERNEL_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&KERNEL_WCET[cpu_id]);
                    write_volatile(&mut KERNEL_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Task => unsafe {
                    let t = read_volatile(&TASK_TIME[cpu_id]);
                    write_volatile(&mut TASK_TIME[cpu_id], t + diff);
                    let c = read_volatile(&TASK_COUNT[cpu_id]);
                    write_volatile(&mut TASK_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&TASK_WCET[cpu_id]);
                    write_volatile(&mut TASK_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Interrupt => unsafe {
                    let t = read_volatile(&INTERRUPT_TIME[cpu_id]);
                    write_volatile(&mut INTERRUPT_TIME[cpu_id], t + diff);
                    let c = read_volatile(&INTERRUPT_COUNT[cpu_id]);
                    write_volatile(&mut INTERRUPT_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&INTERRUPT_WCET[cpu_id]);
                    write_volatile(&mut INTERRUPT_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::ContextSwitch => unsafe {
                    let t = read_volatile(&CONTEXT_SWITCH_TIME[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_TIME[cpu_id], t + diff);
                    let c = read_volatile(&CONTEXT_SWITCH_COUNT[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&CONTEXT_SWITCH_WCET[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::ContextSwitchMain => unsafe {
                    let t = read_volatile(&CONTEXT_SWITCH_MAIN_TIME[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_TIME[cpu_id], t + diff);
                    let c = read_volatile(&CONTEXT_SWITCH_MAIN_COUNT[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&CONTEXT_SWITCH_MAIN_WCET[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Idle => unsafe {
                    let t = read_volatile(&IDLE_TIME[cpu_id]);
                    write_volatile(&mut IDLE_TIME[cpu_id], t + diff);
                    let c = read_volatile(&IDLE_COUNT[cpu_id]);
                    write_volatile(&mut IDLE_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&IDLE_WCET[cpu_id]);
                    write_volatile(&mut IDLE_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Boot => (),
            }
        }

        let cnt = awkernel_lib::delay::cpu_counter();

        unsafe {
            // Overhead of this.
            let t = read_volatile(&PERF_TIME[cpu_id]);
            write_volatile(&mut PERF_TIME[cpu_id], t + (cnt - end));
            let c = read_volatile(&PERF_COUNT[cpu_id]);
            write_volatile(&mut PERF_COUNT[cpu_id], c + 1);
            let wcet = read_volatile(&PERF_WCET[cpu_id]);
            write_volatile(&mut PERF_WCET[cpu_id], wcet.max(cnt - end));

            // State transition.
            write_volatile(&mut START_TIME[cpu_id], cnt);
            write_volatile(&mut PERF_STATES[cpu_id], next_state as u8);
        }
    }

    fn update_time_and_state_for_dag(next_state: PerfState) {Expand commentComment on line R1022
        let end = awkernel_lib::delay::cpu_counter();
        let cpu_id = awkernel_lib::cpu::cpu_id();

        let state: PerfState = unsafe { read_volatile(&PERF_STATES[cpu_id]) }.into();
        if state == next_state {
            return;
        }

        let start = unsafe { read_volatile(&START_TIME[cpu_id]) };

        if start > 0 && start <= end {
            let diff = end - start;

            match state {
                PerfState::Kernel => unsafe {
                    let t = read_volatile(&KERNEL_TIME[cpu_id]);
                    write_volatile(&mut KERNEL_TIME[cpu_id], t + diff);
                    let c = read_volatile(&KERNEL_COUNT[cpu_id]);
                    write_volatile(&mut KERNEL_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&KERNEL_WCET[cpu_id]);
                    write_volatile(&mut KERNEL_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Task => unsafe {
                    let t = read_volatile(&TASK_TIME[cpu_id]);
                    write_volatile(&mut TASK_TIME[cpu_id], t + diff);
                    let c = read_volatile(&TASK_COUNT[cpu_id]);
                    write_volatile(&mut TASK_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&TASK_WCET[cpu_id]);
                    write_volatile(&mut TASK_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Interrupt => unsafe {
                    let t = read_volatile(&INTERRUPT_TIME[cpu_id]);
                    write_volatile(&mut INTERRUPT_TIME[cpu_id], t + diff);
                    let c = read_volatile(&INTERRUPT_COUNT[cpu_id]);
                    write_volatile(&mut INTERRUPT_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&INTERRUPT_WCET[cpu_id]);
                    write_volatile(&mut INTERRUPT_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::ContextSwitch => unsafe {
                    let t = read_volatile(&CONTEXT_SWITCH_TIME[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_TIME[cpu_id], t + diff);
                    let c = read_volatile(&CONTEXT_SWITCH_COUNT[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&CONTEXT_SWITCH_WCET[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::ContextSwitchMain => unsafe {
                    let t = read_volatile(&CONTEXT_SWITCH_MAIN_TIME[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_TIME[cpu_id], t + diff);
                    let c = read_volatile(&CONTEXT_SWITCH_MAIN_COUNT[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&CONTEXT_SWITCH_MAIN_WCET[cpu_id]);
                    write_volatile(&mut CONTEXT_SWITCH_MAIN_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Idle => unsafe {
                    let t = read_volatile(&IDLE_TIME[cpu_id]);
                    write_volatile(&mut IDLE_TIME[cpu_id], t + diff);
                    let c = read_volatile(&IDLE_COUNT[cpu_id]);
                    write_volatile(&mut IDLE_COUNT[cpu_id], c + 1);
                    let wcet = read_volatile(&IDLE_WCET[cpu_id]);
                    write_volatile(&mut IDLE_WCET[cpu_id], wcet.max(diff));
                },
                PerfState::Boot => (),
            }
        }

        let cnt = awkernel_lib::delay::cpu_counter();

        unsafe {
            // Overhead of this.
            let t = read_volatile(&PERF_TIME[cpu_id]);
            write_volatile(&mut PERF_TIME[cpu_id], t + (cnt - end));
            let c = read_volatile(&PERF_COUNT[cpu_id]);
            write_volatile(&mut PERF_COUNT[cpu_id], c + 1);
            let wcet = read_volatile(&PERF_WCET[cpu_id]);
            write_volatile(&mut PERF_WCET[cpu_id], wcet.max(cnt - end));

            // State transition.
            write_volatile(&mut START_TIME[cpu_id], cnt);
            write_volatile(&mut PERF_STATES[cpu_id], next_state as u8);
        }
    }
```

kobayu858: @nokosaaan 
I’ve already checked this, but please double-check just to be sure.

kernel/Cargo.toml
kobayu858: https://github.com/tier4/awkernel/issues/692

awkernel_async_lib/src/task.rs
nokosaaan: I’ve checked it.

kobayu858 (approved):

nokosaaan:
@kobayu858 
I fixed the starting point of `context_switch` in `perf` at commit de7c27f0cd48cc1b9715646b76fe2f1b73c9d106. 
Since placing it in `/awkernel_lib/src/interrupt.rs` would cause a circular dependency with the `perf` module, I moved it to the beginning of `preemption()`, which is called immediately afterward. 

Additionally, the location where I had initially placed it—as discussed earlier—was within an architecture-dependent processing flow, which could prevent the process from running on architectures other than `x86`; therefore, I determined that the current location is appropriate.

As before, there are discrepancies in processing times between context switches via `preemption` and those via `run_main`—due to factors like early returns—which can result in variations in measurement times. 
However, I believe this issue can be resolved by comparing the two.

In this PR, I have left a comment explaining why I placed the code here, and our plan is to make adjustments as needed if a more optimal location is found in the future.

PR #679, opened 2026-04-19
feat(x86): Add x86_64 power support and wire it into awkernel_shell

ytakano:
## Description

This PR adds an x86_64 power support module and integrates it into the current build and shell environment.

The main changes are:
- add a new `awkernel_lib/src/arch/x86_64/power.rs`
- export and wire the new x86_64 power functionality through `awkernel_lib`
- update `applications/awkernel_shell` to use the new functionality
- update build configuration (`Cargo.toml`, `Makefile`) accordingly

Overall, this PR introduces the initial power-related support on x86_64 and connects it to the existing user-facing shell path so that the functionality can be exercised more easily.

## Related links

- Branch: `ytakano/awkernel:x86_power`
- Upstream repository: `tier4/awkernel`

## How was this PR tested?

- built the updated tree with the modified build configuration
- confirmed that the new x86_64 power module is compiled and linked correctly
- verified that the shell-side integration builds successfully
- shutdown and reboot work correctly on physical and virtual machines

## Notes for reviewers

Please focus on:
- the architecture and API design of `arch/x86_64/power.rs`
- whether the new power-related interfaces fit the existing `awkernel_lib` structure
- whether the shell integration is the right place and shape for exposing this functionality
- whether the build system changes are minimal and appropriate

### DIFF
--- a/awkernel_lib/src/arch/x86_64/power.rs
+++ b/awkernel_lib/src/arch/x86_64/power.rs
@@ -1,7 +1,6 @@
 use super::acpi::AcpiMapper;
 use crate::{
-    delay,
-    interrupt,
+    delay, interrupt,
     mmio::{ReadOnlyOffset, WriteOnlyOffset},
     sync::mutex::{MCSNode, Mutex},
 };
--- a/awkernel_lib/src/arch/x86_64/power.rs
+++ b/awkernel_lib/src/arch/x86_64/power.rs
@@ -24,7 +24,7 @@ use x86_64::{
 };
 
 const SCI_EN: u64 = 1 << 0;
-const SLP_TYP_SHIFT: u16 = 10;
+const SLP_TYP_SHIFT: u32 = 10;
 const SLP_EN: u16 = 1 << 13;
 const SLP_TYP_MASK: u64 = 0x7 << SLP_TYP_SHIFT;
 
@@ -59,8 +59,10 @@ struct AcpiRegister {
 pub fn init(acpi: &AcpiTables<AcpiMapper>) -> Result<(), &'static str> {
     let mut node = MCSNode::new();
     let mut state = POWER_CONTROL.lock(&mut node);
-    if !matches!(*state, PowerControlState::Uninitialized) {
-        return Ok(());
+    match *state {
+        PowerControlState::Uninitialized => {}
+        PowerControlState::Ready(_) => return Ok(()),
+        PowerControlState::Failed(err) => return Err(err),
     }
 
     match init_power_control(acpi) {
@@ -341,11 +343,16 @@ fn parse_aml_table(
 ) -> Result<(), &'static str> {
     let mapping =
         unsafe { handler.map_physical_region::<u8>(table.address, table.length as usize) };
-    let data =
-        unsafe { slice::from_raw_parts(mapping.virtual_start().as_ptr(), mapping.region_length()) };
-    context
-        .parse_table(data)
-        .map_err(|_| "Failed to parse AML table.")
+    let result = {
+        let data = unsafe {
+            slice::from_raw_parts(mapping.virtual_start().as_ptr(), mapping.region_length())
+        };
+        context
+            .parse_table(data)
+            .map_err(|_| "Failed to parse AML table.")
+    };
+    AcpiMapper::unmap_physical_region(&mapping);
+    result
 }
 
 fn parse_sleep_types(value: &AmlValue, context: &AmlContext) -> Result<(u16, u16), &'static str> {
--- a/awkernel_lib/src/arch/x86_64/power.rs
+++ b/awkernel_lib/src/arch/x86_64/power.rs
@@ -1,6 +1,7 @@
 use super::acpi::AcpiMapper;
 use crate::{
     delay,
+    interrupt,
     mmio::{ReadOnlyOffset, WriteOnlyOffset},
     sync::mutex::{MCSNode, Mutex},
 };
@@ -14,7 +15,7 @@ use aml::{
     value::{AmlValue, Args},
     AmlContext, AmlName, DebugVerbosity, Handler,
 };
-use core::{ptr, slice};
+use core::{convert::TryFrom, ptr, slice};
 use x86_64::{
     instructions::{
         hlt,
@@ -29,6 +30,7 @@ const SLP_EN: u16 = 1 << 13;
 const SLP_TYP_MASK: u64 = 0x7 << SLP_TYP_SHIFT;
 
 static POWER_CONTROL: Mutex<PowerControlState> = Mutex::new(PowerControlState::Uninitialized);
+static PCI_CONFIG_LOCK: Mutex<()> = Mutex::new(());
 
 #[derive(Clone, Copy)]
 enum PowerControlState {
@@ -56,6 +58,7 @@ struct AcpiRegister {
     address: u64,
 }
 
+/// Initialize x86_64 power control after `delay::init` has completed.
 pub fn init(acpi: &AcpiTables<AcpiMapper>) -> Result<(), &'static str> {
     let mut node = MCSNode::new();
     let mut state = POWER_CONTROL.lock(&mut node);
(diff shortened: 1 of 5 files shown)


awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,592 @@
+use super::acpi::AcpiMapper;
+use crate::{
+    delay,
+    mmio::{ReadOnlyOffset, WriteOnlyOffset},
+    sync::mutex::{MCSNode, Mutex},
+};
+use acpi::{
+    address::{AccessSize, AddressSpace, GenericAddress},
+    fadt::Fadt,
+    AcpiHandler, AcpiTables,
```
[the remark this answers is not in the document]
ytakano: `AcpiHandler` trait is required by `map_physical_region` and `unmap_physical_region` as follows..

```
pub trait AcpiHandler: Clone {
    /// Given a physical address and a size, map a region of physical memory that contains `T` (note: the passed
    /// size may be larger than `size_of::<T>()`). The address is not neccessarily page-aligned, so the
    /// implementation may need to map more than `size` bytes. The virtual address the region is mapped to does not
    /// matter, as long as it is accessible to `acpi`.
    ///
    /// See the documentation on `PhysicalMapping::new` for an explanation of each field on the `PhysicalMapping`
    /// return type.
    ///
    /// ## Safety
    ///
    /// - `physical_address` must point to a valid `T` in physical memory.
    /// - `size` must be at least `size_of::<T>()`.
    unsafe fn map_physical_region<T>(&self, physical_address: usize, size: usize) -> PhysicalMapping<Self, T>;

    /// Unmap the given physical mapping. This is called when a `PhysicalMapping` is dropped, you should **not** manually call this.
    ///
    /// Note: A reference to the handler used to construct `region` can be acquired by calling [`PhysicalMapping::handler`].
    fn unmap_physical_region<T>(region: &PhysicalMapping<Self, T>);
}
```

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    fn read_pci_u8(&self, segment: u16, bus: u8, device: u8, function: u8, offset: u16) -> u8 {
+        (read_pci_config(segment, bus, device, function, offset & !3)
+            >> (((offset & 3) * 8) as u32)) as u8
+    }
+
+    fn read_pci_u16(&self, segment: u16, bus: u8, device: u8, function: u8, offset: u16) -> u16 {
+        (read_pci_config(segment, bus, device, function, offset & !3)
+            >> (((offset & 2) * 8) as u32)) as u16
```
atsushi421, line 520: `read_pci_u16` computes its shift with `(offset & 2) * 8`, correctly assuming 2-byte alignment. But `write_pci_u16` delegates to `write_pci_partial`, which uses `(offset & 3) * 8` for the shift. If an odd offset is ever passed, the `0xFFFF` mask will span a DWORD boundary and corrupt an adjacent byte. Consider either validating alignment in `write_pci_u16` or using `offset & !1` before calling `write_pci_partial`.

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+        Ok(control) => control,
+        Err(err) => {
+            log::error!("Shutdown is unavailable. {err}");
+            qemu_shutdown_fallback();
+        }
+    };
+
+    if let Err(err) = enable_acpi(&control) {
+        log::warn!("Failed to enable ACPI mode before shutdown. {err}");
+    }
```
atsushi421, line 96: Neither `shutdown()` nor `reboot()` disables interrupts before writing to ACPI control registers or legacy reset ports. On SMP systems, an interrupt arriving between PM1a and PM1b writes (or during the legacy reset sequence) could corrupt register state. The ACPI spec recommends disabling interrupts before entering a sleep state. Consider calling `x86_64::instructions::interrupts::disable()` at the top of both functions.

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    let address = 0x8000_0000u32
+        | ((bus as u32) << 16)
+        | ((device as u32) << 11)
+        | ((function as u32) << 8)
+        | ((offset as u32) & 0xFC);
+
+    unsafe {
+        PortWriteOnly::<u32>::new(0xCF8).write(address);
```
atsushi421, line 613: `read_pci_config` and `write_pci_config` perform a two-step write-to-0xCF8 / access-0xCFC sequence without any lock or interrupt guard. A concurrent PCI config access (from another CPU or an interrupt handler) between the two port operations would clobber the address register. `write_pci_partial` compounds this with a non-atomic read-modify-write across three separate port accesses. Consider wrapping these in `without_interrupts` or protecting with a spinlock.

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+            .map_err(|_| "Failed to locate PM1a control block.")?,
+        physical_memory_offset,
+    )?;
+    let pm1b_control = fadt
+        .pm1b_control_block()
+        .map_err(|_| "Failed to locate PM1b control block.")?
+        .map(|register| convert_register(register, physical_memory_offset))
+        .transpose()?;
+
+    let (slp_typa, slp_typb) = parse_s5_sleep_types(acpi, fadt.handler().clone())?;
```
atsushi421, line 160: `fadt.smi_cmd_port` is `u32` but cast to `u16` without a range check. `convert_register` validates that I/O addresses fit in 16 bits, but this field skips that validation. Consider using `u16::try_from(fadt.smi_cmd_port).map_err(...)` for consistency.

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+            log::error!("Failed to write PM1b control block. {err}");
+            qemu_shutdown_fallback();
+        }
+    }
+
+    wait_for_power_transition();
+}
+
+pub fn reboot() -> ! {
+    if let Ok(control) = current_power_control() {
```
atsushi421, line 117: `shutdown()` calls `enable_acpi()` before writing to PM1 control registers, but `reboot()` writes to the ACPI reset register without first ensuring ACPI mode is active. On systems still in legacy mode, the reset register write might be ignored. Consider calling `enable_acpi(&control)` here as well.

awkernel_lib/src/arch/x86_64/power.rs
```diff
@@ -0,0 +1,599 @@
...
+    address_space: AddressSpace,
+    width_bytes: u8,
+    address: u64,
+}
+
+pub fn init(acpi: &AcpiTables<AcpiMapper>) -> Result<(), &'static str> {
+    let mut node = MCSNode::new();
+    let mut state = POWER_CONTROL.lock(&mut node);
+    match *state {
+        PowerControlState::Uninitialized => {}
```
atsushi421, line 65: The three public functions (`init`, `shutdown`, `reboot`) have no doc comments. Key information missing includes: `init` must be called after `delay::init`; `shutdown` uses ACPI S5 with a QEMU fallback; `reboot` uses the ACPI reset register with legacy port fallback. Adding `///` doc comments would help callers understand ordering constraints and fallback behavior.

awkernel_lib/src/arch/x86_64/power.rs
ytakano: Now, `read_pci_config_raw` is called `read_pci_config` with a Mutex lock, which disables interrupts..

awkernel_lib/src/arch/x86_64/power.rs
ytakano: Did `let offset = offset & !1` before this.

awkernel_lib/src/arch/x86_64/power.rs
ytakano: `enable_acpi(&control)` is called before writing to registers.

awkernel_lib/src/arch/x86_64/power.rs
ytakano: Interrupts are disabled first, now.

awkernel_lib/src/arch/x86_64/power.rs
ytakano: Add documentation comments. Thank you.

awkernel_lib/src/arch/x86_64/power.rs
ytakano: `try_from` is used for type cast.

atsushi421 (approved):

Issue #335, opened 2025-03-14
Support Intel 2.5 GbE Controller I225/I226


ytakano:
https://github.com/openbsd/src/blob/master/sys/dev/pci/if_igc.c

PR #680, opened 2026-04-20
fix(igc): stabilize i225 phy bring-up

ytakano:
## Description

Align the Awkernel I225 bring-up path more closely with the BSD igc drivers and improve diagnostics around the PHY reset path.

This change makes the MAC reset path tolerant of auto-read completion timeouts, initializes and synchronizes the MDIC PHY address, powers the PHY up before reading its ID, and clears GO_LINKD/ULP before attempting the hardware reset. The PHY reset path now retries once, logs the relevant control/status registers, and treats a missing RST_COMPL bit as non-fatal when direct MDIC reads show that the PHY is already responding.

On the driver side, tick-based link polling was added so link state can advance even when only the periodic path runs, and link transitions now log speed/duplex changes. In hardware testing over a serial console, the previous hard timeout message was replaced by a benign "PHY reset completion bit did not assert, but PHY responded" diagnostic, and the port reached "Up (Full Duplex)" during boot.

## Related links

## How was this PR tested?

- Tested on a physical machine.

### DIFF
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -251,11 +251,12 @@ impl Igc {
         Ok(igc)
     }
 
-    fn intr(&self, _irq: Option<u16>) -> Result<(), IgcDriverErr> {
+    fn intr(&self, irq: Option<u16>) -> Result<(), IgcDriverErr> {
         // TODO: Handle Tx/Rx interrupts.
 
         let mut inner = self.inner.read();
         let igc_icr = read_reg(&inner.info, igc_regs::IGC_ICR)?;
+        let should_poll_link = irq.is_none() && (igc_icr & igc_defines::IGC_ICR_LSC) == 0;
 
         if (igc_icr & igc_defines::IGC_ICR_LSC) != 0 {
             // Link status change interrupt.
@@ -263,6 +264,13 @@ impl Igc {
             {
                 let mut inner = self.inner.write();
                 inner.igc_intr_link()?;
+            }
+            inner = self.inner.read();
+        } else if should_poll_link {
+            drop(inner);
+            {
+                let mut inner = self.inner.write();
+                inner.igc_poll_link()?;
             }
             inner = self.inner.read();
         }
@@ -802,6 +810,12 @@ impl IgcInner {
         )
     }
 
+    #[inline(always)]
+    fn igc_poll_link(&mut self) -> Result<(), IgcDriverErr> {
+        self.hw.mac.get_link_status = true;
+        self.igc_intr_link()
+    }
+
     fn igc_iff(&mut self) -> Result<(), IgcDriverErr> {
         use igc_regs::*;
 
@@ -1154,6 +1168,8 @@ fn igc_update_link_status(
     hw: &mut IgcHw,
     link_info: &mut LinkInfo,
 ) -> Result<(), IgcDriverErr> {
+    let previous_status = link_info.link_status;
+
     if hw.mac.get_link_status {
         ops.check_for_link(info, hw)?;
     }
@@ -1164,6 +1180,7 @@ fn igc_update_link_status(
             link_info.link_speed = Some(speed);
             link_info.link_duplex = Some(duplex);
             link_info.link_active = true;
+            log::debug!("igc: link up: speed={speed:?}, duplex={duplex:?}");
         }
 
         if link_info.link_duplex == Some(IgcDuplex::Full) {
@@ -1176,9 +1193,14 @@ fn igc_update_link_status(
             link_info.link_speed = None;
             link_info.link_duplex = None;
             link_info.link_active = false;
+            log::debug!("igc: link down");
         }
         LinkStatus::Down
     };
+
+    if previous_status != link_info.link_status {
+        log::info!("igc: link status changed: {}", link_info.link_status);
+    }
 
     Ok(())
 }
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -256,8 +256,6 @@ impl Igc {
 
         let mut inner = self.inner.read();
         let igc_icr = read_reg(&inner.info, igc_regs::IGC_ICR)?;
-        let should_poll_link = irq.is_none() && (igc_icr & igc_defines::IGC_ICR_LSC) == 0;
-
         if (igc_icr & igc_defines::IGC_ICR_LSC) != 0 {
             // Link status change interrupt.
             drop(inner);
@@ -266,7 +264,7 @@ impl Igc {
                 inner.igc_intr_link()?;
             }
             inner = self.inner.read();
-        } else if should_poll_link {
+        } else if irq.is_none() {
             drop(inner);
             {
                 let mut inner = self.inner.write();
@@ -810,6 +808,10 @@ impl IgcInner {
         )
     }
 
+    /// Polls link status when called from a tick-driven path (i.e. `intr` is
+    /// invoked without a real IRQ).  Forces `get_link_status = true` so that
+    /// `igc_intr_link` re-checks the hardware even if a link-change interrupt
+    /// was not observed.
     #[inline(always)]
     fn igc_poll_link(&mut self) -> Result<(), IgcDriverErr> {
         self.hw.mac.get_link_status = true;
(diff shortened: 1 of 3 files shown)


awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
```diff
@@ -277,29 +299,53 @@ pub(super) fn igc_phy_hw_reset_generic(
...
-        write_reg(info, IGC_CTRL, ctrl)?;
-        write_flush(info)?;
+            wait_microsec(150);
 
-        wait_microsec(150);
+            for _ in 0..100000 {
+                phpm = read_reg(info, IGC_I225_PHPM)?;
+                wait_microsec(1);
+                if phpm & IGC_I225_PHPM_RST_COMPL != 0 {
+                    return Ok(());
```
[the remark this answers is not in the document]
ytakano: fixed.

awkernel_drivers/src/pcie/intel/igc/i225.rs
```diff
@@ -544,6 +548,13 @@ fn igc_init_phy_params_i225(
 
     phy.autoneg_mask = AUTONEG_ADVERTISE_SPEED_DEFAULT_2500;
     phy.reset_delay_us = 100;
+    let phy_addr = igc_sync_mdic_phy_addr(info, hw, 1)?;
+
+    let mut phpm = read_reg(info, IGC_I225_PHPM)?;
+    phpm &= !(IGC_I225_PHPM_GO_LINKD | 0x0400);
+    write_reg(info, IGC_I225_PHPM, phpm)?;
```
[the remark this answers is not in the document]
ytakano: fixed

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
```diff
@@ -277,29 +302,65 @@ pub(super) fn igc_phy_hw_reset_generic(
...
+            );
+            return Ok(());
+        }
+
+        let status = read_reg(info, IGC_STATUS).unwrap_or(0);
+        log::debug!(
+            "Timeout expired after a phy reset: ctrl={ctrl:#010x}, status={status:#010x}, phpm={phpm:#010x}"
+        );
 
         Ok(())
```
atsushi421, line 370: When both reset attempts time out and the MDIC fallback reads also fail, this returns `Ok(())` even though the PHY is unresponsive. Subsequent PHY initialization (`igc_get_phy_id`, etc.) will operate on a non-functional PHY, leading to confusing downstream failures. Consider returning an error here, or at minimum elevating the log to `warn` level.

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
```diff
@@ -30,10 +32,30 @@ const IGC_I225_PHPM_DIS_1000: u32 = 0x0040; // Disable 1G globally
...
+    let phy_addr = (mdicnfg & IGC_MDICNFG_PHY_MASK) >> IGC_MDICNFG_PHY_SHIFT;
+    let phy_addr = if phy_addr == 0 {
+        default_addr
+    } else {
+        phy_addr
+    };
+
+    let mdicnfg = (mdicnfg & !IGC_MDICNFG_PHY_MASK) | (phy_addr << IGC_MDICNFG_PHY_SHIFT);
+    write_reg(info, IGC_MDICNFG, mdicnfg)?;
+    hw.phy.addr = phy_addr;
```
atsushi421, line 59: Missing `write_flush` after writing `IGC_MDICNFG`. The PHY reset path in the same file consistently flushes after register writes. Without a flush here, subsequent MDIC operations could execute before the PHY address configuration is committed to hardware.

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
```diff
@@ -30,10 +32,30 @@ const IGC_I225_PHPM_DIS_1000: u32 = 0x0040; // Disable 1G globally
 const IGC_I225_PHPM_SPD_B2B_EN: u32 = 0x0080; // Smart Power Down Back2Back
 const IGC_I225_PHPM_RST_COMPL: u32 = 0x0100; // PHY Reset Completed
 const IGC_I225_PHPM_DIS_100_D3: u32 = 0x0200; // Disable 100M in D3
-const IGC_I225_PHPM_ULP: u32 = 0x0400; // Ultra Low-Power Mode
+pub(super) const IGC_I225_PHPM_ULP: u32 = 0x0400; // Ultra Low-Power Mode
 const IGC_I225_PHPM_DIS_2500: u32 = 0x0800; // Disable 2.5G globally
 const IGC_I225_PHPM_DIS_2500_D3: u32 = 0x1000; // Disable 2.5G in D3
 
+pub(super) fn igc_sync_mdic_phy_addr(
+    info: &mut PCIeInfo,
```
atsushi421, line 44: This function would benefit from a doc comment explaining: what it does (reads PHY address from MDICNFG, falls back to `default_addr` when the field is zero, writes the canonical value back), the side-effect on `hw.phy.addr`, and why `default_addr=1` is the standard I225 PHY address.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -802,6 +810,12 @@ impl IgcInner {
         )
     }
 
+    #[inline(always)]
+    fn igc_poll_link(&mut self) -> Result<(), IgcDriverErr> {
```
atsushi421, line 816: A brief doc comment would help here explaining that this method is used for tick-driven link polling (when `intr` is called without a real IRQ), and why it forces `get_link_status = true` before delegating to `igc_intr_link`.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -251,11 +251,12 @@ impl Igc {
         Ok(igc)
     }
 
-    fn intr(&self, _irq: Option<u16>) -> Result<(), IgcDriverErr> {
+    fn intr(&self, irq: Option<u16>) -> Result<(), IgcDriverErr> {
         // TODO: Handle Tx/Rx interrupts.
 
         let mut inner = self.inner.read();
         let igc_icr = read_reg(&inner.info, igc_regs::IGC_ICR)?;
+        let should_poll_link = irq.is_none() && (igc_icr & igc_defines::IGC_ICR_LSC) == 0;
```
atsushi421: By the time the `else if` branch is reached, the preceding `if` has already established that `(igc_icr & IGC_ICR_LSC) == 0`, so that part of the condition is tautologically true. This variable effectively reduces to `irq.is_none()`. Consider simplifying to `else if irq.is_none()` for clarity.

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
ytakano: The original BSD's code does not return error as follows. So, for faithful porting, this does not return error.
However, I make the log level `debug` `warn`.

https://github.com/freebsd/freebsd-src/blob/975e3605ebb15cbaf5a25c1c9d1f51aed41291d0/sys/dev/igc/igc_phy.c#L891

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
ytakano: Fixed. Thank you.

awkernel_drivers/src/pcie/intel/igc/igc_phy.rs
ytakano: Added a doc comment as follows.

```rust
/// Reads the PHY address from MDICNFG, falls back to `default_addr` if th
e
/// field is zero, writes the resolved address back to MDICNFG, and stores
 it
/// in `hw.phy.addr`.  The default value of 1 matches the hardware power-o
n
/// default and is consistent with the BSD igc driver.
pub(super) fn igc_sync_mdic_phy_addr(
```

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Added a doc comment as follows.

```rust
    /// Polls link status when called from a tick-driven path (i.e. `intr`
 is
    /// invoked without a real IRQ).  Forces `get_link_status = true` so t
hat
    /// `igc_intr_link` re-checks the hardware even if a link-change inter
rupt
    /// was not observed.
    #[inline(always)]
    fn igc_poll_link(&mut self) -> Result<(), IgcDriverErr> 
```

awkernel_drivers/src/pcie/intel/igc.rs

atsushi421 (approved):

PR #683, opened 2026-04-22
feat(igc): add single-queue polling datapath

ytakano:
## Description

This PR adds the first usable single-queue polling datapath for the Intel I225/I226 driver.

It enables the PCIe bus-mastering required for DMA, sets up single-queue TX/RX descriptor rings, and wires the minimal polling path through send(), recv(), can_send(), and capabilities(). It also adds TX reclamation based on TDH/TDT and RX consume/refill handling for the single-queue path.

This change is intentionally limited to the minimal end-to-end datapath. It does not add shell debug commands, NetDevice::debug_dump, IRQ RX buffering, multi-queue support, or checksum offload

## How was this PR tested?

- make clippy
- make x86_64 RELEASE=1
- make check_x86_64

Real-machine smoke test:

- Connected over /dev/ttyUSB0 at 38400
- Ran (reboot) from the Awkernel serial shell
- Confirmed the target rebooted and returned to the shell
- Ran (ifconfig) and confirmed interface enumeration, including the igc-backed interfac

## Notes for reviewers

- Review intent is narrow: minimal polling datapath only.
- Files changed are limited to:
    - awkernel_drivers/src/pcie.rs
    - awkernel_drivers/src/pcie/intel/igc.rs
    - awkernel_drivers/src/pcie/intel/igc/igc_defines.rs
- The main runtime-sensitive pieces are:
    - PCIe endpoint/bridge bus-master enablement
    - single-queue TX descriptor submission/reclaim
    - RX descriptor consume/refill
- Observability, IRQ buffering, multi-queue, and checksum offload are intentionally deferred to later PRs.


awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,152 @@ impl IgcInner {
...
+        &self,
+        que_id: usize,
+        ether_frame: net_device::EtherFrameRef,
+    ) -> Result<(), IgcDriverErr> {
+        if que_id != 0 || !self.link_info.link_active || ether_frame.data.len() > TX_BUFFER_SIZE {
+            return Ok(());
+        }
+
+        if ether_frame.vlan.is_some() {
+            return Ok(());
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -1069,11 +1225,7 @@ fn igc_allocate_pci_resources(info: &mut PCIeInfo) -> Result<(Vec<IRQ>, IRQ), PC
 
     let nmsix = nmsix - 1; // Give one vector to events.
 
-    let nqueues = if nmsix > IGC_MAX_VECTORS {
-        IGC_MAX_VECTORS
-    } else {
-        nmsix
-    };
+    let nqueues = core::cmp::min(nmsix, IGC_MAX_VECTORS.min(1));
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,152 @@ impl IgcInner {
...
+        let (status_error, length, vlan) = {
+            let desc = &rx.rx_desc_ring.as_ref()[idx];
+            let status_error = unsafe { desc.wb.upper.status_error };
+            let length = unsafe { desc.wb.upper.length as usize };
+            let vlan = if status_error & IGC_RXD_STAT_VP != 0 {
+                Some(unsafe { desc.wb.upper.vlan })
+            } else {
+                None
+            };
+            (status_error, length, vlan)
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,160 @@ impl IgcInner {
         Ok(())
     }
 
+    fn igc_txeof(&self, que_id: usize, tx: &mut Tx) -> Result<(), IgcDriverErr> {
+        membar_sync();
+        let reg_tdh = read_reg(&self.info, igc_regs::IGC_TDH(que_id))? as usize;
```
atsushi421: `igc_txeof` reads `IGC_TDH` to determine which descriptors have been processed, but Intel's SDM advises against reading the TX head register as it may return stale values during DMA. Since `IGC_TXD_CMD_RS` is already set on each descriptor, consider using the DD bit in each descriptor's write-back status for reclamation instead, which is the approach used by the Linux kernel igc driver.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,160 @@ impl IgcInner {
...
+        &self,
+        que_id: usize,
+        ether_frame: net_device::EtherFrameRef,
+    ) -> Result<(), IgcDriverErr> {
+        if que_id != 0 {
+            return Err(IgcDriverErr::Param);
+        }
+
+        if !self.link_info.link_active {
+            return Ok(());
```
atsushi421, line 898: When `link_active` is false (here) or when the TX ring is full (line 909), `igc_send` returns `Ok(())` without actually transmitting. The caller cannot distinguish a successful send from a silently dropped packet. Consider returning an error in these cases so the caller can decide whether to retry.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -330,13 +334,18 @@ impl NetDevice for Igc {
...
-        // TODO
-        false
+        let inner = self.inner.read();
+        if !inner.if_flags.contains(NetFlags::RUNNING) || !inner.link_info.link_active {
+            return false;
+        }
+
+        let mut node = MCSNode::new();
+        let mut tx = inner.queue_info.que[0].tx.lock(&mut node);
+        inner.igc_txeof(0, &mut tx).is_ok() && tx.igc_desc_unused() > 0
```
atsushi421: `can_send()` acquires the TX lock, issues a memory barrier, reads `IGC_TDH` via MMIO, and runs the `igc_txeof` reclamation loop. In a typical `if can_send() { send() }` polling pattern, all of this work is duplicated since `send()` performs the same operations. Other NIC drivers in this codebase (igb, ixgbe, genet) implement `can_send()` as a lightweight flag check. Consider making this a simple `if_flags`/`link_active` check and relying on `send()` for reclamation.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -74,6 +75,7 @@ const DEFAULT_ITR: u32 = 1000000000 / (MAX_INTS_PER_SEC * 256);
 
 const MAX_FRAME_SIZE: u32 = 9234;
 const RX_BUFFER_SIZE: usize = 4096 * 3;
+const TX_BUFFER_SIZE: usize = 2048;
```
atsushi421, line 78: `TX_BUFFER_SIZE` is 2048 while `RX_BUFFER_SIZE` is 12288 and `MAX_FRAME_SIZE` is 9234. The driver advertises `VLAN_MTU` but any frame over 2048 bytes is rejected by `igc_send`. This creates an asymmetry where the driver can receive frames it cannot retransmit. Either increase the TX buffer to match the supported frame size, or document this intentional limitation.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,160 @@ impl IgcInner {
...
+            &self.info,
+            igc_regs::IGC_TDT(que_id),
+            tx.next_avail_desc as u32,
+        )?;
+        bus_space_barrier(BUS_SPACE_BARRIER_WRITE);
+
+        Ok(())
+    }
+
+    fn igc_recv(&self, que_id: usize) -> Result<Option<net_device::EtherFrameBuf>, IgcDriverErr> {
```
atsushi421, line 951: `can_send()` checks `NetFlags::RUNNING` before proceeding, but `igc_recv` does not. During `igc_stop()`, there is a window between removing the `RUNNING` flag and freeing DMA buffers where `igc_recv` could access hardware that is being reset. Consider adding a `RUNNING` check here for consistency.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -854,6 +861,160 @@ impl IgcInner {
...
+            (ether_frame.data.len() as u32)
+                | IGC_ADVTXD_DTYP_DATA
+                | IGC_TXD_CMD_DEXT
+                | IGC_TXD_CMD_EOP
+                | IGC_TXD_CMD_IFCS
+                | IGC_TXD_CMD_RS,
+        );
+        read.olinfo_status = u32::to_le((ether_frame.data.len() as u32) << IGC_ADVTXD_PAYLEN_SHIFT);
+
+        tx.next_avail_desc += 1;
```
atsushi421: `next_avail_desc` is incremented and the descriptor is written before the `write_reg(IGC_TDT)` call. If `write_reg` fails, the ring's software state is advanced but the hardware TDT is not updated, leaving a zombie descriptor. Consider advancing the index only after the `write_reg` succeeds.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Because, this implementation is polling based approach. It is better to check queue information. This will be removed after initializing interrupts.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Did not add a `RUNNING` check to `igc_recv()` because `igc_stop()` takes the write lock on `inner`, which already serializes teardown against concurrent read-side access.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: This PR only advertises `VLAN_MTU`, not jumbo-frame TX support, so the current scope remains a minimal single-queue datapath. I will fix RX_BUFFER_SIZE at a later PR.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: `send()` returning `Ok(())` on link-down and temporary ring exhaustion to match the existing `igb` / `ixgbe` driver contract used by the current network stack. Packet loss should be handled by network layers such as TCP.

awkernel_drivers/src/pcie/intel/igc.rs

awkernel_drivers/src/pcie/intel/igc.rs

atsushi421 (approved):

PR #685, opened 2026-04-27
feat(igc): add on-demand netdump observability

ytakano:
## Description

This PR adds on-demand IGC observability without changing datapath behavior.

It introduces a default NetDevice::debug_dump() hook, adds awkernel_lib::net::debug_dump_interface(), and wires the shell netdump(interface_id) command to the IGC driver’s existing register and ring dump logic

## Related links

## Notes for reviewers

This PR is intentionally limited to read-only observability. It does not change queueing, DMA, interrupt handling, or RX buffering behavior.

The shell addition is only netdump(interface_id). Write-path helpers such as add_ipv4, arping4, and set_gateway4 are not included here.

### DIFF
--- a/applications/awkernel_shell/Cargo.toml
+++ b/applications/awkernel_shell/Cargo.toml
@@ -9,6 +9,8 @@ edition = "2021"
 blisp = "0.4"
 log = "0.4"
 byteorder = { version = "1.0", default-features = false }
+num-bigint = { version = "0.4", default-features = false }
+num-traits = { version = "0.2", default-features = false }
 
 [dependencies.awkernel_async_lib]
 path = "../../awkernel_async_lib"
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -315,6 +315,10 @@ impl NetDevice for Igc {
             .or(Err(net_device::NetDevError::DeviceError))
     }
 
+    fn debug_dump(&self) {
+        self.inner.read().dump();
+    }
+
     fn add_multicast_addr(&self, addr: &[u8; 6]) -> Result<(), net_device::NetDevError> {
         let mut inner = self.inner.write();
         inner.multicast_addrs.add_addr(*addr);
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -316,7 +316,8 @@ impl NetDevice for Igc {
     }
 
     fn debug_dump(&self) {
-        self.inner.read().dump();
+        let msg = self.inner.read().dump();
+        log::debug!("igc: dump:\r\n{msg}");
     }
 
     fn add_multicast_addr(&self, addr: &[u8; 6]) -> Result<(), net_device::NetDevError> {
@@ -706,7 +707,7 @@ impl IgcInner {
         }
     }
 
-    fn dump(&self) {
+    fn dump(&self) -> alloc::string::String {
         let mut msg = alloc::string::String::new();
 
         msg = format!("BDF: {}\r\n", self.info.get_bdf());
@@ -806,7 +807,7 @@ impl IgcInner {
             msg = format!("{msg}TDBAL{i}: {tdbal:#08x}\r\n");
         }
 
-        log::debug!("igc: dump:\r\n{msg}");
+        msg
     }
 
     #[inline(always)]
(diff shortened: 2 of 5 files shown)


awkernel_lib/src/net.rs
```diff
@@ -428,6 +428,19 @@ pub fn get_interface(interface_id: u64) -> Result<IfStatus, NetManagerError> {
...
+pub fn debug_dump_interface(interface_id: u64) -> Result<(), NetManagerError> {
+    let net_manager = NET_MANAGER.read();
+
+    let if_net = net_manager
+        .interfaces
+        .get(&interface_id)
+        .ok_or(NetManagerError::InvalidInterfaceID)?;
+
+    if_net.net_device.debug_dump();
+
```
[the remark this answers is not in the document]
ytakano: Fixed.

applications/awkernel_shell/Cargo.toml
```diff
@@ -9,6 +9,8 @@ edition = "2021"
 blisp = "0.4"
 log = "0.4"
 byteorder = { version = "1.0", default-features = false }
+num-bigint = { version = "0.4", default-features = false }
```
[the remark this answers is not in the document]
ytakano: There is no `alloc` feature. This PR ignores this review.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -315,6 +315,10 @@ impl NetDevice for Igc {
             .or(Err(net_device::NetDevError::DeviceError))
     }
 
+    fn debug_dump(&self) {
+        self.inner.read().dump();
```
atsushi421: `debug_dump()` holds `self.inner.read()` for the entire dump, which performs many MMIO register reads, an O(N²) `format!("{msg}...")` build, and a console-locked log write. Concurrent `inner.write()` callers (`up`, `down`, `add_multicast_addr`, the LSC/poll-link path in `intr`) block for that whole time, and on writer-preferring RwLocks subsequent `read()` callers in the datapath (`tick`/`recv`/`send`/`can_send`) also stall — which contradicts the PR description's claim of not affecting interrupt/queue behavior.

Consider capturing the small subset of state needed under the lock and releasing it before the format/log work, or exposing the fields actually used (`info`, `hw.mac.addr`) for lock-free access so the dump matches the read-only intent.

awkernel_lib/src/net/net_device.rs
```diff
@@ -220,6 +220,9 @@ pub trait NetDevice {
         Ok(())
     }
 
+    /// Dump device-specific debug state on demand.
+    fn debug_dump(&self) {}
```
atsushi421: Defaulting to an empty body means `(netdump <id>)` silently returns `Ok(())` for every `NetDevice` that hasn't overridden this — `VirtioNet`, `Genet`, `Igb`, `Ixgbe`, `E1000eExample`. The shell user sees no output and no indication the operation was unsupported, which is misleading observability behavior.

Consider returning `Result<(), NetDevError>` with a default of `Err(NetDevError::Unsupported)` so `debug_dump_interface` can surface it, or have the default emit a clear "debug_dump not implemented for <device>" log line via `device_short_name()`.

awkernel_lib/src/net.rs
```diff
@@ -428,6 +428,23 @@ pub fn get_interface(interface_id: u64) -> Result<IfStatus, NetManagerError> {
     Ok(if_status)
 }
 
+pub fn debug_dump_interface(interface_id: u64) -> Result<(), NetManagerError> {
```
atsushi421, line 436: This new public API has no doc comment, while other `pub fn`s in this module (`get_interface`, `up`, `down`, `tick_interface`, ...) document behavior and error conditions. A caller cannot tell where the dump is emitted (the IGC implementation uses `log::debug!`, but that is not part of the contract), what error variants can be returned, or whether it is safe to invoke during active TX/RX.

Please add a `///` comment specifying the output channel, the returned errors (today only `InvalidInterfaceID`), and any timing constraints.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Thank you. Fixed.

awkernel_lib/src/net/net_device.rs
ytakano: Print warning when `debug_dump` is not supported as follows.

`log::warn!("debug_dump not implemented for this device");`

awkernel_lib/src/net.rs
ytakano: Add a doc comment as follows.

```rust
/// Emit debug state for the interface identified by `interface_id` via `log::debug!`.
///
/// Returns [`NetManagerError::InvalidInterfaceID`] if no interface with that ID exists.
/// The NET_MANAGER read lock is held only to look up and clone the device reference;
/// the device dump runs outside that lock.
```

atsushi421 (approved):

PR #686, opened 2026-04-27
feat(sched): add Partitioned EDF scheduler

ytakano:
## Description

This PR adds a **Partitioned EDF (Earliest Deadline First) scheduler** (`SchedulerType::PartitionedEDF(deadline, core)`), which statically assigns tasks to specific CPU cores and applies EDF scheduling independently within each core.

### Key design

- **Per-CPU run queues**: Each core maintains its own `BinaryHeap` ordered by absolute deadline, ensuring scheduling decisions on one core do not affect others.
- **Static task-to-core binding**: Tasks are pinned to a designated core at spawn time via `PartitionedEDF(relative_deadline_us, core_id)`. The assigned core is stored in `Task::partitioned_core`.
- **Intra-core EDF preemption**: When a newly woken task has an earlier deadline than the task currently running on the same core, an IPI is sent to trigger preemption. The preemption target is determined as `max(running_task, pending_preemption)` to avoid unnecessary preemption inversion.
- **Priority over GEDF**: `PartitionedEDF` sits at the top of `PRIORITY_LIST`, above `GEDF`. CPUs with partitioned tasks in queue exclusively serve those tasks.
- **DAG support**: Absolute deadline calculation reuses `gedf::calculate_and_update_dag_deadline` for DAG-aware task graphs.
- **Worker wake-up**: `wake_workers` wakes cores with partitioned tasks via a dedicated loop, separately from the general non-partitioned task distribution.

### Files

| File | Change |
|------|--------|
| `awkernel_async_lib/src/scheduler/partitioned_edf.rs` | New scheduler implementation |
| `awkernel_async_lib/src/scheduler.rs` | Add `PartitionedEDF` variant, `get_num_partitioned_schedulers()` const fn, updated `get_next_task` |
| `awkernel_async_lib/src/task.rs` | Add `Task::partitioned_core`, `NUM_PARTITIONED_TASKS_IN_QUEUE`, `get_task_running`, updated `wake_workers` |
| `applications/tests/test_partitioned_edf/` | Test application (core pinning, EDF preemption, multi-core independence) |

## How was this PR tested?

- **Build**: `make x86_64 RELEASE=1` passes with no errors.
- **Test application** (`test_partitioned_edf`): built and linked via `--features test_partitioned_edf`. Three test scenarios are implemented:
  1. **Core pinning** — each task logs `cpu_id()` and checks it matches the assigned core; mismatch is reported as `[FAIL]`.
  2. **EDF preemption** — `heavy` (deadline=9900ms) and `light` (deadline=990ms) are pinned to core 1; `light` is expected to preempt `heavy` based on EDF ordering, observable from start/end timestamps in the log.
  3. **Multi-core independence** — two tasks on core 1 and core 2 (skipped when `num_cpu < 3`) run in parallel; each task's `cpu_id()` log confirms no cross-core migration.

To run:
```
RUSTFLAGS="-C panic=unwind" cargo +nightly-2025-11-16 x86 --release --features test_partitioned_edf
```

## Notes for reviewers

- `get_num_partitioned_schedulers()` is a `const fn` that counts leading `PartitionedEDF` entries in `PRIORITY_LIST` at compile time. If a new partitioned scheduler type is added, it must be placed at the front of `PRIORITY_LIST` and the `matches!` pattern updated accordingly.
- Core ID 0 (BSP) is intentionally excluded from partitioned task assignment. `wake_workers` skips core 0 (`.skip(1)`), and the task spawn path emits `log::warn!` if `core == 0` or `core >= num_cpu()` is given.
- The preemption logic in `invoke_preemption` compares the new task against `max(running_task, pending_preemption)`, mirroring the GEDF implementation. This prevents unnecessary preemption when a higher-priority task is already pending.
- `NUM_TASK_IN_QUEUE` was changed from `AtomicU64` to `AtomicU32` on 64-bit targets to match the 32-bit path; task counts do not require 64-bit range.


kernel/Cargo.toml
```diff
@@ -135,3 +135,4 @@ aarch64_virt = [
 ]
 spinlock = ["awkernel_lib/spinlock"]
 linux = []
+test_partitioned_edf = ["userland/test_partitioned_edf"]
```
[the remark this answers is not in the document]
ytakano: Remove this from the description.

awkernel_async_lib/src/task.rs
```diff
@@ -299,13 +305,25 @@ impl Tasks {
...
+                let partitioned_core = scheduler_type.partitioned_core();
+                if let Some(i) = partitioned_core {
+                    if i == 0 || i >= num_cpu() as u16 {
+                        log::warn!(
+                            "Partitioned core should be between 1 and {}. It is addressed as {}.",
+                            num_cpu() - 1,
+                            i
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
ytakano: Fallback to core 1 as follows.

```rust
                let partitioned_core = if let Some(i) = scheduler_type.partitioned_core() {
                    if i == 0 || i >= num_cpu() as u16 {
                        log::warn!(
                            "Partitioned core should be between 1 and {}. Falling back to core 1. Given core: {}",
                            num_cpu() - 1,
                            i
                        );
                        Some(1)
                    } else {
                        Some(i)
                    }
                } else {
                    None
                };
```

awkernel_async_lib/src/scheduler/partitioned_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+            NUM_PARTITIONED_TASKS_IN_QUEUE[partitioned_core].fetch_add(1, Ordering::Relaxed);
+
+            let mut node_inner = MCSNode::new();
+            let mut data = self.data[partitioned_core].lock(&mut node_inner);
+            let internal_data = data.get_or_insert_with(EDFData::new);
+            internal_data.queue.push(PartitionedEDFTask {
+                task: task.clone(),
+                absolute_deadline,
+                wake_time,
+            });
```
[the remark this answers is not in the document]
ytakano: Check it as follows.


```Rust
        let partitioned_core = partitioned_core as usize;
        if partitioned_core >= num_cpu() || partitioned_core == 0 {
            panic!("PartitionedEDF: core {partitioned_core} is out of range");
        }
```

awkernel_async_lib/src/scheduler/partitioned_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+        };
+
+        let mut node = MCSNode::new();
+        let _guard = GLOBAL_WAKE_GET_MUTEX.lock(&mut node);
+        if !self.invoke_preemption(task.clone()) {
+            let partitioned_core = partitioned_core as usize;
+            if partitioned_core >= self.data.len() {
+                panic!("PartitionedEDF: core {partitioned_core} exceeds max supported CPU count");
+            }
+
```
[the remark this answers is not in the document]
ytakano: Check this as follows.

```rust
        let partitioned_core = partitioned_core as usize;
        if partitioned_core >= num_cpu() || partitioned_core == 0 {
            panic!("PartitionedEDF: core {partitioned_core} is out of range");
        }
```

awkernel_async_lib/src/scheduler/partitioned_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+                    info.update_absolute_deadline(absolute_deadline);
+
+                    (wake_time, absolute_deadline, partitioned_core)
+                }
+                _ => unreachable!(),
+            }
+        };
+
+        let partitioned_core = partitioned_core as usize;
+        if partitioned_core >= num_cpu() || partitioned_core == 0 {
```
atsushi421, line 96: `Tasks::new` already validates `partitioned_core` and falls back to `Some(1)` with a warning when invalid, but `wake_task` re-reads the original value from `info.scheduler_type` and panics here. The fallback path is therefore dead code: any task spawned with `core == 0` or `core >= num_cpu()` will panic the kernel on first wake despite the friendly-warning intent. Read `task.partitioned_core` (the validated value) here, or rewrite `info.scheduler_type` at spawn time so both paths agree.

awkernel_async_lib/src/task.rs
```diff
@@ -134,7 +138,9 @@ impl ArcWake for Task {
             panicked = info.panicked;
         }
 
-        NUM_TASK_IN_QUEUE.fetch_add(1, Ordering::Release);
+        if self.partitioned_core.is_none() {
```
atsushi421: When a partitioned task panics, the wake path falls into the global `panicked::SCHEDULER`, but this conditional skips the `NUM_TASK_IN_QUEUE` increment because `partitioned_core.is_some()`. The partitioned scheduler's counter is also not touched in this path. Neither counter reflects the queued task, so `wake_workers` will not wake any CPU and the panicked task can sit indefinitely. Always increment `NUM_TASK_IN_QUEUE` for the panicked path.

awkernel_async_lib/src/scheduler/partitioned_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+    },
+};
+use alloc::{collections::BinaryHeap, sync::Arc};
+use awkernel_lib::{
+    cpu::{num_cpu, NUM_MAX_CPU},
+    sync::mutex::{MCSNode, Mutex},
+};
+
+pub struct PartitionedEDFScheduler {
+    data: [Mutex<Option<EDFData>>; NUM_MAX_CPU], // Run queue.
```
atsushi421: `Mutex<Option<EDFData>>` plus `get_or_insert_with(EDFData::new)` adds a branch and an extra `as_mut()` to every `wake_task` / `get_next`. `BinaryHeap::new()` does not allocate until pushed, so the lazy init buys nothing — and the slot is created `NUM_MAX_CPU` times regardless. Use `[Mutex<EDFData>; NUM_MAX_CPU]` (or a `BinaryHeap` directly) and drop the `Option`.

awkernel_async_lib/src/task.rs
```diff
@@ -1125,10 +1152,18 @@ impl Ord for PriorityInfo {
 
 /// Wake workers up.
 pub fn wake_workers() {
+    let num_cpus = awkernel_lib::cpu::num_cpu();
     let mut num_tasks = NUM_TASK_IN_QUEUE.load(Ordering::Relaxed);
-    let num_cpu = awkernel_lib::cpu::num_cpu();
 
-    for i in 1..num_cpu {
+    for (i, partitioned_tasks) in NUM_PARTITIONED_TASKS_IN_QUEUE[..num_cpus]
```
atsushi421, line 1172: `wake_workers` iterates `NUM_PARTITIONED_TASKS_IN_QUEUE[..num_cpus].iter().enumerate().skip(1)` twice — first to skip CPUs that have partitioned work, then to wake those same CPUs. Each atomic load acts as an optimization barrier, doubling memory traffic on a hot path. Fuse into one loop: load the count once, `wake_cpu` and `continue` if `> 0`, otherwise fall through to the global wake logic.

awkernel_async_lib/src/task.rs
```diff
@@ -952,6 +974,11 @@ pub fn get_tasks_running() -> Vec<RunningTask> {
     tasks
 }
 
+pub fn get_task_running(cpu_id: usize) -> RunningTask {
```
atsushi421, line 991: This new public function lacks a doc comment. The convention that `task_id == 0` means "idle" and the valid range of `cpu_id` (`< num_cpu()`) are non-obvious, and the relationship to the existing `get_tasks_running()` is unclear. Add a `///` comment covering return-value semantics and bounds.

awkernel_async_lib/src/scheduler.rs
```diff
@@ -94,28 +100,51 @@ impl SchedulerType {
                 | (SchedulerType::Panicked, SchedulerType::Panicked)
         )
     }
+
+    pub const fn partitioned_core(&self) -> Option<u16> {
```
atsushi421, line 108: This new public method has no doc comment. The semantics of `Some(n)` (the assigned core index, expected to be in `1..num_cpu()`) and the fact that all non-`PartitionedEDF` variants return `None` should be documented for consumers.

awkernel_async_lib/src/scheduler/partitioned_edf.rs
ytakano: Use `task`'s `partitioned_core` as follows.

```rust
        let partitioned_core =
            task.partitioned_core.expect("Task has no partitioned core") as usize;
```

awkernel_async_lib/src/task.rs
ytakano: Fixed. A doc comment has been added:

```rust
/// Get the running task for the given CPU.
///
/// `cpu_id` must be less than `awkernel_lib::cpu::num_cpu()`.
/// The returned [`RunningTask::task_id`] is `0` when the CPU is idle.
///
/// To get the running tasks for all CPUs at once, use [`get_tasks_running`].
```

awkernel_async_lib/src/scheduler.rs
ytakano: Fixed. A doc comment has been added:

```rust
/// Return the partitioned core index if this is a [`SchedulerType::PartitionedEDF`] scheduler.
///
/// Returns `Some(n)` where `n` is the CPU core index (`1..num_cpu()`) assigned to the
/// partitioned EDF scheduler. Returns `None` for all other scheduler variants.
```

awkernel_async_lib/src/task.rs
ytakano: Fixed. The second loop has been removed. The single loop now handles both cases in one pass: load NUM_PARTITIONED_TASKS_IN_QUEUE[i] once per CPU; if it is > 0 call wake_cpu(i) and continue; otherwise fall through to the global task logic in the same iteration.

awkernel_async_lib/src/scheduler/partitioned_edf.rs
ytakano: Fixed. The field is now [Mutex<EDFData>; NUM_MAX_CPU] initialized via const fn EDFData::new() which returns an empty BinaryHeap. Since BinaryHeap::new() does not allocate, there is no cost difference compared to the Option-based lazy init, and the branch and as_mut() are gone from the hot path.

awkernel_async_lib/src/task.rs
ytakano: Fixed. The increment condition in Task::wake() is now:

```rust
if panicked || self.partitioned_core.is_none() {
    NUM_TASK_IN_QUEUE.fetch_add(1, Ordering::Release);
}
```

When panicked is true the counter is always incremented regardless of partitioned_core, so wake_workers will correctly wake a CPU for the panicked task. The panicked task is then routed to panicked::SCHEDULER which reads from NUM_TASK_IN_QUEUE.

atsushi421 (approved):

PR #681, opened 2026-04-20
feat(app): add autoware module (imu_driver and vehicle_velocity_converter)

nokosaaan:
## Description
I have added the `imu_driver` and `vehicle_velocity_converter `modules. While I based the test code on the implementation in `Autoware`, there was no test code for `imu_driver`, so I created test cases focused on input and output. All test cases have been verified.

## Related links
https://github.com/autowarefoundation/autoware_core/tree/main/sensing/autoware_vehicle_velocity_converter/test
https://github.com/tier4/tamagawa_imu_driver/tree/ros2

## How was this PR tested?
cargo test --manifest-path /path/to/module/Cargo.toml
[test_imu_driver.log](https://github.com/user-attachments/files/26891221/test_imu_driver.log)
[test_vehicle_velocity_converter.log](https://github.com/user-attachments/files/26891535/test_vehicle_velocity_converter.log)

PR #687, opened 2026-04-27
fix(app): add autoware module (imu_corrector)

nokosaaan:
## Description
I added the `imu_corrector` module.
Basically, I added it in the same format as `imu_driver` and `vehicle_velocity_converter`, but since I could only find test code for `gyro_bias_estimator` and not for `imu_corrector_core`, I added tests to verify the correctness of the input and output as a substitute, similar to `imu_driver`.

## Related links
https://github.com/autowarefoundation/autoware_universe/tree/main/sensing/autoware_imu_corrector/src

## How was this PR tested?
You should add the path to the Cargo.toml file you want to unit test to the `members` section of the `workspace` in the root directory's Cargo.toml (at `/awkernel/Cargo.toml`)
[imu_corrector.log](https://github.com/user-attachments/files/27125552/imu_corrector.log)


applications/tests/test_autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,478 @@
...
+        UnitQuaternion::from_quaternion(n_quat)
+    }
+
+    fn transform_vector3(&self, vec: &Vector3, transform: &Transform) -> Vector3 {
+        let nalgebra_vec = self.to_nalgebra_vector3(vec);
+        let nalgebra_quat = self.to_nalgebra_quaternion(&transform.rotation);
+        let nalgebra_trans = self.to_nalgebra_vector3(&transform.translation);
+        let rotated = nalgebra_quat * nalgebra_vec;
+        let result = rotated + nalgebra_trans;
+        self.to_imu_vector3(&result)
```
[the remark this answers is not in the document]
nokosaaan: @copilot 
I fixed it on commit id 11537c0a275ec3335c35ad8192c0407f8c9cf955

applications/tests/test_autoware/imu_corrector/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "imu_corrector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+log = "0.4"
+nalgebra = { version = "0.32", default-features = false, features = ["libm"] }
+awkernel_async_lib = { path = "../../../../awkernel_async_lib", default-features = false }
+awkernel_lib = { path = "../../../../awkernel_lib", default-features = false }
```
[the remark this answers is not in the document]
nokosaaan: I fixed it.

applications/autoware/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+path = "src/lib.rs"
+crate-type = ["rlib"]
+
+[dependencies]
+log = "0.4"
+libm = "0.2"
+csv-core = "0.1"
+awkernel_async_lib = { path = "../../../awkernel_async_lib", default-features = false }
+awkernel_lib = { path = "../../../awkernel_lib", default-features = false }
+imu_driver = { path = "./imu_driver", default-features = false }
```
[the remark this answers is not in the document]
nokosaaan: I fixed it on commit id 6d82977a28a4d6f8d15a820efe51f67fbc3a0b65

applications/autoware/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#![no_std]
+extern crate alloc;
+
+pub async fn run() {}
```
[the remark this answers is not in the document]
nokosaaan: No need to fix i t for now.

kobayu858:
This PR will be reviewed after #681 is merged.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
+// Ported from the following versions of the original C++ code:
+// universe/autoware_universe:
+// type: git
+// url: https://github.com/autowarefoundation/autoware_universe.git
+// version: 0.51.0
```
kobayu858, line 21: The attribution comment only references the repository and version, but not the specific source file. Please also specify the original file path. Additionally, the original source is Apache 2.0 licensed; the license header should be carried over.

applications/autoware/imu_driver/src/lib.rs
```diff
@@ -53,6 +53,7 @@
 // original file path: src/tag_serial_driver.cpp
 // test code: Created in-house for I/O verification
 // version: 0.1.0
+// version: 0.1.0
```
kobayu858: remove

applications/autoware/imu_corrector/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "imu_corrector"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+nalgebra = { version = "0.32", default-features = false, features = ["libm"] }
+imu_driver = { path = "../imu_driver", default-features = false }
```
kobayu858: add trailing blank line

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
+// Ported from the following versions of the original C++ code:
```
kobayu858: `correct_for_dynamic_bias` and `correct_for_scale` are not implemented in this port.

Since these are present in the original C++ implementation, could you add a comment in `ImuCorrectorConfig` explaining why they are omitted and marking them as `TODO` for future support?

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            corrected_imu.linear_acceleration_covariance =
+                self.transform_covariance(&corrected_imu.linear_acceleration_covariance);
+            corrected_imu.angular_velocity =
+                self.transform_vector3(&corrected_imu.angular_velocity, tf);
+            corrected_imu.angular_velocity_covariance =
+                self.transform_covariance(&corrected_imu.angular_velocity_covariance);
+            corrected_imu.header.frame_id = self.config.output_frame;
+        }
+
+        corrected_imu
```
kobayu858, line 266: The behavior when a transform is unavailable differs from the original C++ implementation.

In the original, if the TF lookup fails, the function logs an error and **returns without publishing anything**:
```cpp
if (!tf_imu2base_ptr) {
    RCLCPP_ERROR(this->get_logger(), "Please publish TF %s to %s", ...);
    return;
}
```

In the Rust version, `correct_imu_with_covariance` accepts `transform: Option<&Transform>`, and when `None` is passed, it **returns the uncorrected data as-is** (without frame conversion). This means uncorrected IMU data in `imu_link` frame could be published as if it were in `base_link`, which may cause incorrect behavior in downstream modules such as EKF.

Please add a comment explaining why you chose this setting.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+            angular_velocity_covariance: [0.0; 9],
+            linear_acceleration: imu_msg.linear_acceleration.clone(),
+            linear_acceleration_covariance: [0.0; 9],
+        }
+    }
+
+    pub fn to_imu_msg(&self) -> ImuMsg {
+        ImuMsg {
+            header: self.header.clone(),
+            orientation: self.orientation.clone(),
```
kobayu858: In `ImuWithCovariance::from_imu_msg`, the `orientation` from the original IMU message is copied as-is:
```rust
orientation: imu_msg.orientation.clone(),
```

However, the original C++ implementation intentionally does **not** copy `orientation` into the output message after the frame transform:
```cpp
sensor_msgs::msg::Imu imu_msg_base_link;
imu_msg_base_link.linear_acceleration = transform_vector3(...);
imu_msg_base_link.angular_velocity = transform_vector3(...);
// ← no assignment to orientation
```

Please add a comment explaining why you chose this setting.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+    fn to_imu_vector3(&self, vec: &NVector3<f64>) -> Vector3 {
+        Vector3::new(vec.x, vec.y, vec.z)
+    }
+
+    fn to_nalgebra_quaternion(&self, quat: &Quaternion) -> UnitQuaternion<f64> {
+        let n_quat = NQuaternion::new(quat.w, quat.x, quat.y, quat.z);
+        UnitQuaternion::from_quaternion(n_quat)
+    }
+
+    pub fn apply_to_vector(&self, vec: Vector3) -> Vector3 {
```
kobayu858, line 50: Is this unused code? If you plan to use it later, please leave a comment.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    fn transform_vector3(&self, vec: &Vector3, transform: &Transform) -> Vector3 {
+        let nalgebra_vec = self.to_nalgebra_vector3(vec);
+        let nalgebra_quat = self.to_nalgebra_quaternion(&transform.rotation);
+        let rotated = nalgebra_quat * nalgebra_vec;
+        self.to_imu_vector3(&rotated)
+    }
+
+    fn transform_covariance(&self, cov: &[f64; 9]) -> [f64; 9] {
```
kobayu858: The `ImuCorrector::transform_covariance` (private method) and `transform_covariance` (public function) use the same logic. Please either consolidate them into a single implementation or refactor the method so that it internally calls the public function.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,488 @@
...
+        assert_eq!(transformed_cov[1], 0.0);
+        assert_eq!(transformed_cov[2], 0.0);
+        assert_eq!(transformed_cov[3], 0.0);
+        assert_eq!(transformed_cov[5], 0.0);
+        assert_eq!(transformed_cov[6], 0.0);
+        assert_eq!(transformed_cov[7], 0.0);
+    }
+
+    #[test]
+    fn public_covariance_transform_uses_max_diagonal() {
```
kobayu858: You are testing the private method and the public function separately, even though they have the same input and the same expected output. Please consolidate the tests after eliminating duplication and resolving the issue of duplicate definitions.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_driver/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_corrector/Cargo.toml
nokosaaan: I fixed it.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it and added TODO comment.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I have added a comment to the code reflecting the details below.

**① Awkernel DAG Architectural Constraints**

Since C++’s `callback_imu` is called independently as a ROS2 subscription callback, the pipeline does not stop even if it `returns`. In contrast, Awkernel DAG imposes a type constraint requiring reactors to always return an `OutputTuple`; if no output is provided, the downstream `gyro_odometer` (which waits for two inputs) will be permanently blocked. “Returning without publishing” cannot be implemented in this architecture.

**② The `tf` value used for evaluation was fixed and was set to a value that would not cause any issues even if it were `None`.**

It is empirically known that the `tf` value obtained by `imu_corrector` returns the same value every time. We speculate that the issue did not manifest because the `tf` value used for evaluation had a rotation of identity (unit quaternion), so even if `None` was passed as an argument, it did not cause a problem. In fact, in the subsequent [autoware/src/lib.rs](https://github.com/tier4/awkernel/blob/d1c287582c27c31bb442c4a6fda5a5e4077120ab/applications/autoware/src/lib.rs#L201), `None` is passed as the argument for the transform in `correct_imu_with_covariance` within the `imu_corrector reactor`.
However, considering future evaluations in various scenarios (dynamic `tf`, real world), the current implementation regarding the `tf` value is vulnerable, and at the very least, a fixed `tf` value should be used in the correction calculations. Therefore, in a subsequent PR, we plan to hardcode a fixed `tf` value to prevent None from being passed as an argument. We do not believe this change will alter the results of previous evaluations.

For an explanation of why TF is fixed, please refer to [this link](https://azumi-lab2022.slack.com/archives/C054R6DL5KL/p1756107258103419?thread_ts=1755854741.411639&cid=C054R6DL5KL) (This was discussed in the Slack group "group-awkernel").
If you cannot view it, please refer to the link below.

> Hi, thanks for your help the other day. I was a bit confused then, so I wanted to double-check this. I'm looking at the code here: https://github.com/autowarefoundation/autoware_universe/blob/4fdc88454d6cab9d150d6c0bdd142317849d553e/sensing/autoware_imu_corrector/src/imu_corrector_core.cpp#L107. Am I correct that this value is a fixed one? And if I'm right, it's a value that's set in the YAML file for each product, something like this, correct?
> https://github.com/tier4/aip_launcher/blob/5036f3d575385b44259d4f26c9c5ceaeb23a1cc1/aip_x2_gen2_description/config/sensors_calibration.yaml#L75
> (Seems like *_description directory is actually used in runtime not just documenting?)
> 
> Hi!:raised_hand_with_fingers_splayed:
> Yest that is the correct TF and its fixed, so if you want to test without tf_get_latest_transform you only need to consider that TF and because is fixed it should not be a problem, the tf is published as static so it will not change the TF

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I added a comment.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: This code is used as part of the `gyro_odometer` function processing.
(Since this is tf-related processing, it currently performs no action; however, as noted in other comments, it will be necessary for a fix in a subsequent PR.)

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+            rotation: imu_driver::Quaternion {
+                x: 0.0,
+                y: 0.0,
+                z: 0.0,
+                w: 1.0,
+            },
+        }
+    }
+
+    fn to_nalgebra_vector3(&self, vec: &Vector3) -> NVector3<f64> {
```
kobayu858: to_nalgebra_vector3, to_imu_vector3, and to_nalgebra_quaternion all perform pure type conversions without using any fields of self, so there is no need for them to be &self instance methods.
Additionally, identical implementations exist in ImuCorrector, creating a risk of the two diverging if only one is updated in the future.
Please consider refactoring — for example, by extracting these as free functions or implementing the standard From/Into traits.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+        timestamp: u64,
+    ) -> Option<Transform>;
+}
+
+pub struct MockTransformListener {
+    transforms: alloc::collections::BTreeMap<String, Transform>,
+}
+
+impl MockTransformListener {
+    pub fn new() -> Self {
```
kobayu858, line 75: Please add an implementation for the Default trait.

applications/autoware/imu_corrector/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+    }
+}
+
+pub struct ImuCorrector<T: TransformListener> {
+    config: ImuCorrectorConfig,
+    transform_listener: T,
+}
+
+impl ImuCorrector<MockTransformListener> {
+    pub fn new() -> Self {
```
kobayu858, line 175: Please add an implementation for the Default trait.

applications/autoware/imu_corrector/src/lib.rs
kobayu858: If orientation isn't being used in the first place, wouldn't it make sense to just remove it? Is there a clear reason to keep it around? Either way, I don't think the current code will ever actually be used as-is in the future. What do you think?

nokosaaan: I included the `orientation` field in the `ros_bag` used for evaluation because it contained that data, and I assumed that if there was an input, that value would naturally be included in the output. However, as you pointed out, `orientation` is not actually included in the actual processing, so I plan to remove it as a member of the struct.

nokosaaan: In this case, I realized that I would also need to modify `imu_driver`, which is where `ImuMsg` is defined.

nokosaaan: Upon checking the data from the `imu_driver` in `ros_bag`, I found that during simulation, the `orientation` values were fixed at `x:0.0, y:0.0, z:0.0, w:1.0`. These values represent an identity rotation, which aligns with the IMU’s design intent of not affecting the `orientation`. Therefore, I’m considering a policy where, while the sensor data includes `orientation`, we do not include it in the input/output messages (i.e., we remove orientation from `ImuMsg` as well). What do you think?

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I extracted them as free functions.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_corrector/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/imu_corrector/src/lib.rs
kobayu858: Agreed. Since orientation is always identity in simulation and is not used by any downstream node, removing it from ImuMsg is the right call. Keeping unused fields in a message type only adds noise. The change to imu_driver looks clean as well.

kobayu858 (approved):

PR #688, opened 2026-05-08
feat(app): add autoware module (gyro_odometer)

nokosaaan:
## Description
I added the `gyro_odometer` module.
I have ported the Autoware test cases to Rust code.
I consolidated the header structures and other types previously defined separately in `imu_driver` and `vehicle_velocity_converter` into a common type and defined them in a separate file (`common_types`).

## Related links
https://github.com/autowarefoundation/autoware_core/tree/main/localization/autoware_gyro_odometer/test

## How was this PR tested?
[gyro_odometer.txt](https://github.com/user-attachments/files/27574328/gyro_odometer.txt)


kobayu858:
This PR will be reviewed after #687 is merged.

applications/autoware/gyro_odometer/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+            }
+
+            if Self::check_timeout(current_time, latest_imu_stamp, self.message_timeout_sec) {
+                self.vehicle_twist_queue.clear();
+                self.gyro_queue.clear();
+                return Err(GyroOdometerError::TimeoutError(String::from(
+                    "IMU message timeout",
+                )));
+            }
+        }
```
kobayu858: The timeout check timing in `concat_gyro_and_odometer` differs from the original C++ implementation. The current implementation only performs the timeout check when both queues are non-empty. In the original C++, the last received timestamp is managed in a field independent of the queue (`latest_vehicle_twist_ros_time_`), and the timeout check is performed before the queue empty check. As a result, if one queue is empty and messages from the other source stop arriving, the timeout will not be detected. For example, even if IMU messages have been absent for 2 seconds, it will go unnoticed as long as `gyro_queue` is empty. Please fix this.

applications/autoware/gyro_odometer/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+
+        let tf = self.get_transform(
+            &self.gyro_queue.front().unwrap().header.frame_id,
+            &self.output_frame,
+        )?;
+
+        for gyro in &mut self.gyro_queue {
+            let transformed_angular_velocity = tf.apply_to_vector(gyro.angular_velocity.clone());
+            gyro.angular_velocity = transformed_angular_velocity;
+        }
```
kobayu858, line 131: https://github.com/tier4/awkernel/pull/687#discussion_r3280818718

In the original C++ implementation, `angular_velocity_covariance` should also be transformed via `transform_covariance()` after the angular_velocity vector transformation. This is currently omitted because the TF used in evaluation is a fixed identity transform, making the result identical. Please add this as a comment in the code.

applications/autoware/gyro_odometer/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+    pub fn get_transform(&self, from_frame: &str, to_frame: &str) -> Result<Transform> {
+        if from_frame == to_frame || from_frame == "" || to_frame == "" {
+            Ok(Transform::identity())
+        } else {
+            Ok(Transform::identity())
+        }
+    }
+
+    pub fn process_result(
```
kobayu858, line 220: Regarding the discrepancy between C++'s `publish_data` (which outputs 4 topics: `TwistStamped` raw, `TwistWithCovarianceStamped` raw, `TwistStamped` corrected, and `TwistWithCovarianceStamped` corrected) and the current `process_result` which returns only a single `TwistWithCovarianceStamped`: is this intentional because the other three outputs are not required in the minimal MRM DAG configuration? If so, please add a comment to the code stating that the remaining outputs are omitted as they are not needed in the current minimal MRM DAG configuration.

kobayu858 (changes requested):

applications/autoware/gyro_odometer/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+    pub fn check_timeout(current_timestamp: u64, last_timestamp: u64, timeout_sec: f64) -> bool {
+        let dt = (current_timestamp as f64 - last_timestamp as f64) / 1_000_000_000.0;
+        dt.abs() > timeout_sec
+    }
+    pub fn get_transform(&self, from_frame: &str, to_frame: &str) -> Result<Transform> {
+        if from_frame == to_frame || from_frame == "" || to_frame == "" {
+            Ok(Transform::identity())
+        } else {
+            Ok(Transform::identity())
+        }
```
kobayu858, line 212: https://github.com/tier4/awkernel/pull/687#discussion_r3280818718

This is understood given the architectural constraints. However, please leave a comment in the code noting that, in the original implementation, the queues should be cleared and the function should return when the transform lookup fails.

applications/autoware/gyro_odometer/src/lib.rs
nokosaaan: I fixed it.

applications/autoware/gyro_odometer/src/lib.rs
nokosaaan: I added a comment.

applications/autoware/gyro_odometer/src/lib.rs
nokosaaan: I added a comment.

applications/autoware/gyro_odometer/src/lib.rs
nokosaaan: I added a comment.

kobayu858 (approved):

PR #689, opened 2026-05-08
feat(app): add autoware module (ekf_localizer)

nokosaaan:
### Summary

Ports `autoware_ekf_localizer`'s (autoware_core 1.8.0) `EKFModule` (`ekf_module.cpp`) and its
dependencies (`common/autoware_kalman_filter`, `state_transition.cpp`, `measurement.cpp`,
`mahalanobis.cpp`, `covariance.cpp`, `numeric.hpp`) to a new `ekf_localizer` Rust crate for
awkernel. Resolves the following 5 gaps/simplifications flagged in review:

1. Missing `measurement_update_pose`
2. Simplified `measurement_update_twist` (no message covariance, no Mahalanobis gate)
3. Missing low-speed gate (`vx_obs_var` was hardcoded)
4. Missing delay-compensation buffer (extended state)
5. `initialize()` not TF-aware

### Module layout vs. upstream

```mermaid
graph TD
    subgraph crate["ekf_localizer crate"]
        lib["lib.rs (EKFModule / EKFParameters)"]
        kf["kalman_filter.rs (DelayCompensatedKalmanFilter)"]
        st["state_transition.rs"]
        meas["measurement.rs"]
        maha["mahalanobis.rs"]
        cov["covariance.rs"]
        num["numeric.rs"]
        warn["warn_throttle.rs"]
    end
    lib --> kf
    lib --> st
    lib --> meas
    lib --> maha
    lib --> cov
    lib --> num
    lib --> warn

    subgraph upstream["upstream C++ (autoware_core 1.8.0)"]
        u_ekf["ekf_module.cpp / .hpp"]
        u_kf["common/autoware_kalman_filter (time_delay_kalman_filter.cpp)"]
        u_st["state_transition.cpp"]
        u_meas["measurement.cpp"]
        u_maha["mahalanobis.cpp"]
        u_cov["covariance.cpp"]
        u_num["numeric.hpp"]
        u_warn["ROS 2 rcutils (RCUTILS_LOG_CONDITION_THROTTLE_BEFORE)"]
    end
    lib -.port of.-> u_ekf
    kf -.port of.-> u_kf
    st -.port of.-> u_st
    meas -.port of.-> u_meas
    maha -.port of.-> u_maha
    cov -.port of.-> u_cov
    num -.port of.-> u_num
    warn -.algorithm ported from.-> u_warn
```

```mermaid
graph LR
    pose["PoseWithCovarianceStamped (NDT/GNSS, etc.)"] --> mup["measurement_update_pose"]
    twist["TwistWithCovarianceStamped (vehicle_velocity_converter + gyro_odometer)"] --> gate["apply_twist_observability_gate (inflates vx variance at low speed)"]
    gate --> mut["measurement_update_twist"]
    mup --> kfstate["DelayCompensatedKalmanFilter (extended state x_ex / P_ex)"]
    mut --> kfstate
    kfstate --> getters["get_current_pose / get_current_twist / ..."]
```

### Deliberately left out of this port

| Item | Where in upstream | Why omitted |
|---|---|---|
| `EKFDiagnosticInfo` (diagnostics accumulation) | `ekf_module.hpp` | awkernel has no `diagnostic_updater`-equivalent sink yet (YAGNI). `measurement_update_pose/twist` simply return `bool` |
| `diagnostics.*` parameters (`pose/twist_no_update_count_threshold_*`, `ellipse_scale`, `error/warn_ellipse_size*`, `diagnostics_publish_frequency`, etc.) | `hyper_parameters.hpp` / yaml | Excluded for the same reason, since they only configure the diagnostics above |
| `node.show_debug_info`, `node.predict_frequency` (→ `ekf_rate`/`ekf_dt`), `node.tf_rate` | same | ROS node-layer settings (timer frequency, debug prints); irrelevant to `EKFModule`'s own logic |
| `pose_measurement.max_pose_queue_size`, `twist_measurement.max_twist_queue_size` | same | pub/sub queueing config, belongs to the node layer (`ekf_localizer.cpp`), out of `EKFModule`'s scope |
| An `is_mrm_mode`/`set_mrm_mode` flag | (added, then removed during this work) | Switching into MRM/Dead Reckoning is a pub/sub-layer concern realized purely by no longer publishing `pose_with_covariance` -- it should not live as internal state on `EKFModule`. The earlier draft incorrectly gated twist updates on it too |
| The simplified `update_velocity(vx, wz)` (no-covariance) API | (an earlier simplification) | Unified into `measurement_update_twist`. Choosing between vx (vehicle) and wz (IMU gyro) sources is already resolved by `gyro_odometer`, which fuses them into one `TwistWithCovarianceStamped` before this crate ever sees it |
| The `Warning` class / `warning_message.cpp` itself | `src/warning_message.cpp` etc. | Designed around holding a node's own `rclcpp::Clock`, which has no awkernel equivalent. Replaced with `log::warn!` + a per-call-site `WarnThrottle` (see below) |
| `test_aged_object_queue.cpp`, `test_diagnostics*.cpp`, `test_warning_message.cpp`, `test_string.cpp` | `test/` | The corresponding implementations (diagnostics, `AgedObjectQueue`, the `Warning` class's own string handling) were not ported, so porting only their tests would be meaningless |
| The two Python launch/integration tests | `test/` | Assume ROS 2 launch files and real/simulated system integration; awkernel has no equivalent execution environment |
| `TimeDelayKalmanFilterTest.UpdateWithNegativeDelayStep` | `test_time_delay_kalman_filter.cpp` | Upstream's `delay_step` is a signed `int`; this crate uses `delay_step: usize`, so a negative value is simply not constructible. The type system enforces this at compile time, making the corresponding runtime test unnecessary |

### On `log::warn!`

Upstream's `Warning::warn_throttle(msg, duration_ms)` calls ROS 2's `RCLCPP_WARN_THROTTLE` macro, which expands into `rcutils`'s `RCUTILS_LOG_CONDITION_THROTTLE_BEFORE` macro. Checking the actual
source (`rclcpp`/`rcutils`, `humble` branch) showed this to be a simple mechanism: a `static`
variable pair per call site (`last_logged`, initialized to `0`, plus `duration`), throttled by the
condition `now >= last_logged + duration`.

awkernel has no equivalent ROS node or `rclcpp::Clock`, so this mechanism couldn't be ported as-is. `warn_throttle.rs` introduces a `WarnThrottle` struct instead, and `EKFModule` holds one independent instance per call site (8 in total, across pose/twist × frame_id/delay_time/delay_step/mahalanobis). This reproduces upstream's property that a sensor failing one gate every tick doesn't also suppress warnings from a different gate.

**awkernel's own `log::warn!` (`buffered_logger.rs`) allocates via `alloc::format!`, takes an MCS
mutex, and does a non-blocking channel push on every call -- it has no throttling built in.**
Because of this, `WarnThrottle::should_emit()` is checked *before* calling `log::warn!`, unlike
upstream, which always builds the message string first regardless of whether the throttle will
actually let it through (so the formatting/allocation cost is paid on every call there, even whilethrottled). Since `measurement_update_pose/twist` sit on the EKF's RT-critical per-tick update path, this ordering matters under a sustained sensor fault.

Upstream's two separate `warn_throttle` calls on the Mahalanobis gate (the distance value, and a fixed "Ignore the measurement data." message) are combined into a single `log::warn!` call in the Rust port (no information lost, just fewer allocations).

`WarnThrottle` itself was initially implemented with two bugs -- "always emit on the first call"
and "fail open on a backward time jump" -- both found and corrected after reading the actual `rcutils` macro source:
- First call: `last_logged` starts at `0` (not "unset"), so the first call is also suppressed if
  `now < duration`.
- Backward time jump: `now < last_logged` makes the condition `false`, so upstream fails *closed*  (suppresses). A fail-open fallback was the wrong design.

## Related links
https://github.com/autowarefoundation/autoware_core/tree/1.8.0/localization/autoware_ekf_localizer/src
https://github.com/autowarefoundation/autoware_core/tree/1.8.0/localization/autoware_ekf_localizer/test
https://github.com/autowarefoundation/autoware_core/tree/1.8.0/common/autoware_kalman_filter
https://github.com/ros2/geometry2/blob/rolling/tf2/include/tf2/LinearMath/Matrix3x3.hpp
https://github.com/ros2/geometry2/tree/humble/tf2/include/tf2/impl
https://github.com/ros2/common_interfaces/tree/rolling/geometry_msgs/msg

## How was this PR tested?
[ekf_localizer.txt](https://github.com/user-attachments/files/30608947/ekf_localizer.txt)

### Intent behind crate-only tests

Most already carry a `// [own test] no upstream equivalent; ...`-style comment added at the time, so see the source for details. Worth calling out at a larger scale:

- **`kalman_filter.rs`'s ground-truth cross-check suite** (the `ground_truth` module,
  `GroundTruthFixture`, and 5 cross-check tests): ports the exact methodology upstream's own  `test_time_delay_kalman_filter.cpp` uses -- an independently reimplemented, non-optimized  `ground_truth_predict`/`ground_truth_update` pair checked against the real (sparsity-optimized)  implementation. Only this `dim_x=3` cross-check can catch a transposed block or wrong slice offset; the `dim_x=1` tests elsewhere cannot, since every "matrix" there is a 1x1 scalar.
- **All of `lib.rs`'s tests**: upstream has no `test_ekf_module.cpp` at this tag -- `EKFModule`
  itself is never unit-tested upstream -- so every test exercising `EKFModule`'s behavior here is this crate's own integration-level test. All 13 test functions are marked with `[own test]`.
- **`warn_throttle.rs`'s 5 tests**: the throttle algorithm itself belongs to ROS 2's `rcutils`, not
  `autoware_ekf_localizer` (and `rcutils`'s own unit tests are out of scope for this repo), so these verify this crate's implementation against the macro's documented semantics instead.

### Notable fidelity fixes found during this review

- `quaternion_to_yaw`/`quaternion_to_rpy`: fetching the actual `tf2::getYaw`/
  `tf2::Matrix3x3::getEulerYPR` source (`ros2/geometry2`, `humble` branch) showed the "normal case" formulas already matched, but the gimbal-lock fallback branch (pitch near +/-90 deg) was missing and has been added.
- `find_closest_delay_time_index`: replaced a linear scan with `slice::partition_point` (an actual binary search, matching `std::lower_bound`'s algorithm class, and available in `core` so it stays `no_std`-safe).
- `nanos_to_seconds_delta`: switched from converting each `u64` to `f64` before subtracting, to subtracting in exact `u64` nanoseconds first and converting the (much smaller) difference to `f64` afterward -- both more precise, and closer to how `rclcpp::Time`/`Duration` keep nanoseconds as an integer internally.


kobayu858:
This PR will be reviewed after https://github.com/tier4/awkernel/pull/688 is merged.

kobayu858:
@nokosaaan 
The following five items appear to be missing. If this is intentional, please specify so in the comments; if they were simply overlooked, please implement them.

1. `measurement_update_pose` does not exist.
2. `measurement_update_twist` has been simplified.
3. `vx_obs_var` is fixed at 1.0, causing the logic that “does not trust velocity observations at low speeds” to be lost.
4.  The delay compensation buffer is not functioning.
5. The initialization (initialize) does not account for TF transformations.
     Is it perhaps assuming that the input `ros2bag` is already transformed?
     
Are you implementing only “predictions for a single point in time + simplified velocity observation updates”?

nokosaaan:
@kobayu858 
You're right — without these 5, this isn't a faithful port of upstream `ekf_localizer`, and I want to implement all 5 rather than leave them as documented gaps: `measurement_update_pose`, the full `measurement_update_twist` (message-covariance + Mahalanobis gate), the low-speed `vx_obs_var gate`, the delay-compensation extended state, and TF-aware `initialize()`.

On whether the current state is intentional: it reflects what was used for our paper's evaluation (dead-reckoning only, `imu_raw.csv + velocity_status.csv (from awkernel_script)`, no pose observation input). I checked whether any of the 5 were a deliberate "minimal MRM set" decision. Answering the two I was most unsure about:

**3. vx_obs_var fixed at 1.0 — not intentional, and not dead code.**

- The `threshold_observable_velocity_mps` gate itself (the low-speed on/off switch in upstream's [callback_twist_with_covariance](https://github.com/autowarefoundation/autoware_core/blob/main/localization/autoware_ekf_localizer/src/ekf_localizer.cpp#L366-L370)) wasn't exercised in our comparison: the param file we used for the reference run sets it to `0.0`, matching upstream's own [default](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/localization/ekf_localizer.param.yaml#L66) ("Set 0.0 if you want to ignore"). So I can't claim the gate's absence caused a measurable divergence in that specific comparison — both sides had it off.
- But [vx_obs_var = 1.0](https://github.com/tier4/awkernel/blob/ekf_localizer/applications/autoware/ekf_localizer/src/lib.rs#L366) itself — the value `update_velocity` uses unconditionally regardless of that gate — doesn't trace back to anything. It doesn't match the value the rest of our own pipeline actually uses: [VehicleVelocityConverter](https://github.com/tier4/awkernel/blob/main/applications/autoware/vehicle_velocity_converter/src/lib.rs#L105) defaults `stddev_vx` to `0.2`, which [create_covariance_matrix](https://github.com/tier4/awkernel/blob/main/applications/autoware/vehicle_velocity_converter/src/lib.rs#L142) turns into a variance of `0.04` on the published twist message — not `1.0`. So `1.0` looks like an unreviewed placeholder rather than an intentional choice, independent of whether the low-speed gate is on.
- It's also not unreachable: the actual `test_autoware` DAG wiring calls [ekf.update_velocity(vx, wz)](https://github.com/tier4/awkernel/blob/test_autoware/applications/autoware/src/lib.rs#L290-L292) on every `50ms` tick, fed by  `vehicle_velocity_converter→ gyro_odometer` over the same CSV data (`5700 rows, ~25% below 0.05 m/s`).

**5. `initialize()` missing TF — looked fine only because the harness, not the code, kept both sides aligned.**

- Our own comparison setup aligns the Autoware and Awkernel runs to start from the same initial pose as an input-alignment step, and the DAG's [pose_dummy_generator](https://github.com/tier4/awkernel/blob/test_autoware/applications/autoware/src/lib.rs#L244-L267) feeds a hardcoded `(0,0,0)` / identity-quaternion Pose — not anything derived from a real pose source or TF. So the comparison never actually exercised a real map-frame/sensor-mount offset; it isn't a validated decision that TF-handling is safe to omit, just an artifact of how the demo seeds its starting pose.

Given all that, I'll implement all 5 to match upstream, independent of whether today's specific evaluation setup happens to exercise each one.

PR #691, opened 2026-05-15
fix(igc): buffer rx packets on irq

ytakano:
This is PR4 in the igc incremental stack (stacked on top of PR3 / igc_observability; PR1–3 are already merged).

## PR4 changes (this commit)

Decouples RX packet delivery from immediate upper-layer consumption by
introducing a software ring queue drained during interrupt handling.

Key modifications to `awkernel_drivers/src/pcie/intel/igc.rs`:
- Adds `read_queue: RingQ<EtherFrameBuf>` (32-packet capacity) to `Rx`
- Replaces `igc_recv()` with `igc_rx_recv(que_id, rx: &mut Rx)` that
- `recv()` fast-path pops from `read_queue`; falls back to `igc_rx_recv()`
- `intr()` calls `igc_rx_recv()` on RX IRQ, buffering packets before upper-layer consumption
- Resets `read_queue` in `igc_allocate_queues()`, `igc_setup_rx_desc_ring()`, and `igc_stop()`

`applications/tests/test_network` adjusted: `INTERFACE_ID` → 1, addresses → 192.168.100.x, tests narrowed to `tcp_listen_test` + new `udp_recv_test`.

## PR4 review follow-up (also in this commit)

Addressed Copilot review feedback:
- Fixed no-op `self.dump()` in `igc_up()`: captures and logs the returned String via `log::debug!` (regression introduced during PR3 refactoring when `dump()` was changed to return String instead of logging directly)
- Changed `let _ = rx.read_queue.push(packet)` to increment `dropped_pkts` on failure, making any unexpected push failure observable via the existing drop counter (the `while !is_full()` guard makes this path unreachable in practice)
- Removed "via `log::debug!`" from `debug_dump_interface()` doc comment; the log level is device-defined
- Updated `(netdump if)` help text to `(netdump id)` to avoid the keyword-like `if`

Not adopted: kept `warn!` in the default `NetDevice::debug_dump()` (PR3 design decision — operators should see a clear message when calling `netdump` on a device that has not implemented the hook); did not alter the PR
description for `debug_dump`/`netdump` additions (those are PR3 changes visible because PR4 is stacked on PR3).

## Deferred to later PRs

- TX reclaim IRQ path (PR5+): `can_send()` remains a reclaiming gate
- Multi-queue support (PR5)
- TX checksum offload (PR6)

https://github.com/tier4/awkernel/issues/335

## Testing

- `make clippy` — clean
- `make x86_64 RELEASE=1` — successful
- `make check_x86_64` — passed
[...]

PR #693, opened 2026-06-01
feat(igc): enable multi-queue tx/rx

ytakano:
## Description

This PR adds multi-queue TX/RX support to the IGC driver. It is the fifth step in the incremental IGC bring-up stack; PR4 (#691) must be merged before this one.

## Related links

- PR1 igc_phy_bringup — #680
- PR2 igc_single_queue_polling — #683
- PR3 igc_observability — #685
- PR4 igc_irq_rx_buffering — #691 (base of this PR)
- https://github.com/tier4/awkernel/issues/335

## How was this PR tested?

- [x] make x86_64 RELEASE=1 — build passes
- [x] make check_x86_64 — unit tests pass
- [x] Real-machine validation

## Notes for reviewers

- igc_txeof ignores _que_id because the DD-bit reclaim implementation (adopted in PR4's review) does not need the hardware head register. The parameter is present only for call-site uniformity and future extensibility.
- Queue count is capped at 4 to match the I225's RSS queue limit. The actual count will be 1 on systems where num_cpu() returns 1 or where few MSI-X vectors are available, so single-queue regression risk is low.

### DIFF
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -1271,10 +1271,7 @@ fn igc_is_valid_ether_addr(addr: &[u8; 6]) -> bool {
 }
 
 fn igc_select_num_queues(available_vectors: usize) -> usize {
-    let cpu_count = match awkernel_lib::cpu::num_cpu() {
-        0 => 4,
-        n => n,
-    };
+    let cpu_count = core::cmp::max(1, awkernel_lib::cpu::num_cpu());
     let available = core::cmp::min(core::cmp::min(available_vectors, cpu_count), 4);
 
     if available >= 4 {
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -279,8 +279,15 @@ impl Igc {
         let igc_icr = read_reg(&inner.info, igc_regs::IGC_ICR)?;
         let irq_queue = irq.and_then(|irq| inner.queue_info.irqs_to_queues.get(&irq).copied());
 
+        // Accumulate errors rather than short-circuiting with `?`: the EIMS/IMS re-arm
+        // writes below must execute regardless of queue-service failures, otherwise the
+        // interrupt line stays masked permanently.
+        let mut result: Result<(), IgcDriverErr> = Ok(());
+
         if let Some(que_id) = irq_queue {
-            Self::service_queue(&inner, que_id)?;
+            if let Err(e) = Self::service_queue(&inner, que_id) {
+                result = Err(e);
+            }
         }
 
         let should_poll_link = irq.is_none() && (igc_icr & igc_defines::IGC_ICR_LSC) == 0;
@@ -290,21 +297,27 @@ impl Igc {
             drop(inner);
             {
                 let mut inner = self.inner.write();
-                inner.igc_intr_link()?;
+                if let Err(e) = inner.igc_intr_link() {
+                    result = Err(e);
+                }
             }
             inner = self.inner.read();
         } else if should_poll_link {
             drop(inner);
             {
                 let mut inner = self.inner.write();
-                inner.igc_poll_link()?;
+                if let Err(e) = inner.igc_poll_link() {
+                    result = Err(e);
+                }
             }
             inner = self.inner.read();
         }
 
         if irq.is_none() {
             for que_id in 0..inner.queue_info.que.len() {
-                Self::service_queue(&inner, que_id)?;
+                if let Err(e) = Self::service_queue(&inner, que_id) {
+                    result = Err(e);
+                }
             }
         }
 
@@ -317,7 +330,7 @@ impl Igc {
             msix_queuesmask | msix_linkmask,
         )?;
 
-        Ok(())
+        result
     }
 }
 
@@ -379,8 +392,8 @@ impl NetDevice for Igc {
 
         for que_id in 0..inner.queue_info.que.len() {
             let mut node = MCSNode::new();
-            let mut tx = inner.queue_info.que[que_id].tx.lock(&mut node);
-            if inner.igc_txeof(que_id, &mut tx).is_ok() && tx.igc_desc_unused() > 0 {
+            let tx = inner.queue_info.que[que_id].tx.lock(&mut node);
+            if tx.igc_desc_unused() > 0 {
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -265,21 +279,11 @@ impl Igc {
...
-                inner.igc_txeof(&mut tx)
-            };
-            rx_result.and(tx_result)
-        } else {
-            Ok(())
-        };
+        if let Some(que_id) = irq_queue {
+            Self::service_queue(&inner, que_id)?;
+        }
+
```
atsushi421, line 292: Using `?` here short-circuits before the `IGC_IMS` / `IGC_EIMS` re-arm writes below. A transient error from `igc_rx_recv` or `igc_txeof` will leave queue/link interrupts masked and the NIC silent. Capture the first error in a local, continue servicing the remaining queues, and return it only after the EIMS/IMS writes have executed — matching the prior `que_result` semantics.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -925,10 +943,6 @@ impl IgcInner {
         que_id: usize,
```
atsushi421, line 959: The previous `if que_id != 0 { return Err(Param) }` guard is removed without a replacement bounds check, so `self.queue_info.que[que_id]` will panic for any out-of-range caller. Add `if que_id >= self.queue_info.que.len() { return Err(IgcDriverErr::Param); }` at the top of `igc_send`, and consider the same for `igc_rx_recv` (which also lost its `debug_assert_eq!` guard).

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -365,9 +377,15 @@ impl NetDevice for Igc {
             return false;
         }
 
-        let mut node = MCSNode::new();
-        let mut tx = inner.queue_info.que[0].tx.lock(&mut node);
-        inner.igc_txeof(&mut tx).is_ok() && tx.igc_desc_unused() > 0
+        for que_id in 0..inner.queue_info.que.len() {
+            let mut node = MCSNode::new();
+            let mut tx = inner.queue_info.que[que_id].tx.lock(&mut node);
+            if inner.igc_txeof(que_id, &mut tx).is_ok() && tx.igc_desc_unused() > 0 {
```
atsushi421: `igc_txeof` performs a `membar_sync()` plus a descriptor walk; calling it for every queue on the `can_send` probe path serializes TX locks and contends with the ISR-side `service_queue`. Since the descriptor reclaim is already driven by the interrupt/poll path, `tx.igc_desc_unused() > 0` alone is enough here — drop the `igc_txeof` call from this hot path.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -1258,6 +1270,19 @@ fn igc_is_valid_ether_addr(addr: &[u8; 6]) -> bool {
     !(addr[0] & 1 != 0 || addr.iter().all(|&x| x == 0))
 }
 
+fn igc_select_num_queues(available_vectors: usize) -> usize {
+    let cpu_count = core::cmp::max(1, awkernel_lib::cpu::num_cpu());
+    let available = core::cmp::min(core::cmp::min(available_vectors, cpu_count), 4);
+
+    if available >= 4 {
```
atsushi421, line 1311: After clamping by `min(available_vectors, cpu_count, 4)`, the result is snapped to `{1, 2, 4}`, so 3 available vectors collapse to 2 queues and waste a vector/CPU. The EIMS mask and `igc_initialize_rss_mapping` both handle arbitrary `nqueues`, so the power-of-two stepping is unnecessary — return the clamped value directly. If a power-of-two is required by some RSS constraint, please add a comment stating that constraint.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -1258,6 +1270,19 @@ fn igc_is_valid_ether_addr(addr: &[u8; 6]) -> bool {
     !(addr[0] & 1 != 0 || addr.iter().all(|&x| x == 0))
 }
 
+fn igc_select_num_queues(available_vectors: usize) -> usize {
```
atsushi421, line 1307: This function encodes a non-trivial policy (MSI-X vector count, CPU count, hard cap of 4, power-of-two snap) but has no doc comment. The `4` in particular is the I225 RSS queue limit — without a `///` doc it reads as a magic number once the PR description is gone. Please add a doc comment explaining the three constraints and the rationale for the cap.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -896,7 +914,7 @@ impl IgcInner {
         Ok(())
     }
 
-    fn igc_txeof(&self, tx: &mut Tx) -> Result<(), IgcDriverErr> {
+    fn igc_txeof(&self, _que_id: usize, tx: &mut Tx) -> Result<(), IgcDriverErr> {
```
atsushi421, line 933: The reason `_que_id` is unused (DD-bit reclaim does not need the hardware head register, parameter kept for call-site uniformity / future per-queue stats) lives only in the PR body. A future reader seeing the underscore will wonder whether this is an oversight. Please move that rationale into a short comment above the function.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Added a doc comment. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Added a doc comment. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: It is required for RSS. Added a doc comment. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs

atsushi421 (approved):

PR #694, opened 2026-06-08
feat(igc): enable tx checksum offload

ytakano:
## Description

- Add ActiveChecksumContext enum to Tx to track and reuse context descriptors
- Add igc_tx_ctx_setup() to program the Advanced TX Context Descriptor for IPv4, TCP/IPv4, and UDP/IPv4 checksum offload
- Update igc_send() to prepend a context descriptor when offload context changes
- Advertise CSUM_IPv4, CSUM_TCPv4, and CSUM_UDPv4 capabilities
- Enable TCP checksum offload in if_net and update the capability comment

## Related links


awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -954,6 +971,104 @@ impl IgcInner {
...
+                if ether_frame
+                    .csum_flags
+                    .contains(PacketHeaderFlags::IPV4_CSUM_OUT)
+                {
+                    olinfo_status |= IGC_TXD_POPTS_IXSM << IGC_ADVTXD_POPTS_SHIFT;
+                    offload = true;
+                }
+                type_tucmd_mlhl |= IGC_ADVTXD_TUCMD_IPV4;
+                ip.header_len() as u32
+            }
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -954,6 +971,104 @@ impl IgcInner {
...
+        }
+
+        // Write a new context descriptor at head.
+        type_tucmd_mlhl |= IGC_ADVTXD_DTYP_CTXT | IGC_TXD_CMD_DEXT;
+        let mss_l4len_idx = l4len << IGC_ADVTXD_L4LEN_SHIFT;
+
+        let desc = &mut tx.tx_desc_ring.as_mut()[head];
+        desc.adv_ctx.vlan_macip_lens = u32::to_le(vlan_macip_lens);
+        desc.adv_ctx.ts.launch_time = u32::to_le(0);
+        desc.adv_ctx.type_tucmd_mlhl = u32::to_le(type_tucmd_mlhl);
```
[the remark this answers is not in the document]
ytakano: This is incorrect. ussafe is required.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -979,24 +1094,27 @@ impl IgcInner {
...
-        if tx.igc_desc_unused() == 0 {
+        let head = tx.next_avail_desc;
+        let ring_len = tx.tx_desc_ring.as_ref().len();
+
+        let (ctx_count, data_olinfo_status) = self.igc_tx_ctx_setup(&mut tx, &ether_frame, head)?;
+
+        let needed = ctx_count + 1;
+        if tx.igc_desc_unused() < needed {
             return Ok(());
         }
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -114,11 +119,20 @@ struct Rx {
     dropped_pkts: u64,
 }
 
+#[derive(Debug, PartialEq)]
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+            Ok(e) => e,
+            Err(_) => return Ok((0, base_olinfo, None)),
+        };
+
+        let mut olinfo_status = base_olinfo;
+        let mut type_tucmd_mlhl = 0u32;
+        let mut vlan_macip_lens = 0u32;
+        let mut offload = false;
+
+        vlan_macip_lens |= (core::mem::size_of::<EtherHeader>() as u32) << IGC_ADVTXD_MACLEN_SHIFT;
```
atsushi421: Inline 802.1Q VLAN tags pass through `extract_headers()` and are still classified as IPv4/TCP/UDP, but MACLEN is hardcoded to `size_of::<EtherHeader>()` (14) and `l4_off` uses `ETHER_HDR_LEN`. For a frame carrying an 802.1Q tag the real L2 header is 18 bytes, so the NIC computes the L3/L4 checksum over the wrong range and the pseudo-header seed lands at the wrong offset.

Suggest deriving the L2 length from `ext.ether_vlan.is_some()` (14 vs 18) and using it for both `vlan_macip_lens` MACLEN and `l4_off`.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
-            }
+            tx.next_to_clean = if idx + 1 == ring_len { 0 } else { idx + 1 };
         }
 
         Ok(())
     }
 
+    /// Compute the IPv4 pseudo-header checksum that must be seeded into the L4
+    /// checksum field so the NIC can complete TCP/UDP checksum offload.
+    fn igc_pseudo_cksum(ip: &Ip) -> u16 {
```
atsushi421, line 1004: `igc_pseudo_cksum()` and the seed offset computation assume the L4 header lives at `ETHER_HDR_LEN + iphlen`, but for any IP fragment (non-zero fragment offset or MF=1) the L4 header is not there. The seed is written into payload bytes and the NIC then "completes" a checksum over the wrong field, producing broken packets on the wire.

Suggest bailing out of offload when `(ip.ip_off.swap_bytes() & 0x3fff) != 0` and falling back to the software path.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+        };
+
+        let mut olinfo_status = base_olinfo;
+        let mut type_tucmd_mlhl = 0u32;
+        let mut vlan_macip_lens = 0u32;
+        let mut offload = false;
+
+        vlan_macip_lens |= (core::mem::size_of::<EtherHeader>() as u32) << IGC_ADVTXD_MACLEN_SHIFT;
+
+        let NetworkHdr::Ipv4(ip) = &ext.network else {
```
atsushi421, line 1042: `tx_packet_header_flags()` sets `TCP_CSUM_OUT`/`UDP_CSUM_OUT` for any TCP/UDP transport regardless of network family, but the new capability is `CSUM_TCPv4`/`CSUM_UDPv4` only and this function early-returns for non-IPv4. An IPv6+TCP/UDP packet thus bypasses both smoltcp's software checksum (skipped because `cap.checksum.{tcp,udp}.tx() == false`) and the NIC offload, and leaves the wire with a zero/garbage L4 checksum.

Either gate the flags on IPv4 in `tx_packet_header_flags()`, or split the capability bits per family in `if_net.rs`.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+        &self,
+        tx: &mut Tx,
+        ether_frame: &net_device::EtherFrameRef,
+        head: usize,
+    ) -> Result<(usize, u32, Option<L4CksumSeed>), IgcDriverErr> {
+        let base_olinfo = (ether_frame.data.len() as u32) << IGC_ADVTXD_PAYLEN_SHIFT;
+
+        let ext = match extract_headers(ether_frame.data) {
+            Ok(e) => e,
+            Err(_) => return Ok((0, base_olinfo, None)),
```
atsushi421: If `extract_headers()` returns `Err`, this returns `(0, base_olinfo, None)` and the data descriptor is posted without offload. But smoltcp's TX layer, observing `cap.checksum.{tcp,udp}.tx() == false`, did not compute the L4 checksum either, so the frame goes out with an unfilled checksum field.

Treat this as a hard error (e.g. `Err(IgcDriverErr::InvalidPacket)`) and drop the packet, mirroring how `igb`/`ixgbe` handle it.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+            let done = u32::from_le(unsafe { tx.tx_desc_ring.as_ref()[idx].wb.status })
+                & IGC_TXD_STAT_DD
+                != 0;
             if !done {
-                break;
+                // Context descriptors carry no DD writeback. A context descriptor is
+                // always immediately followed by its data descriptor, and a single queue
+                // completes in order, so if the next descriptor is done this one is a
+                // consumed context descriptor and can be reclaimed; otherwise stop.
+                let next = if idx + 1 == ring_len { 0 } else { idx + 1 };
```
atsushi421, line 977: The look-ahead lacks an `idx == tx.next_avail_desc` guard. When the ring is empty (`next_to_clean == next_avail_desc`), if the next slot still has a stale DD bit set (e.g. on a fresh queue before the first reclaim cycle), the loop treats `idx` as a consumed context descriptor and advances `next_to_clean` past `next_avail_desc`, corrupting the `igc_desc_unused()` accounting.

Add `if idx == tx.next_avail_desc { break; }` at the top of the loop, and guard the look-ahead so it only fires when `next != tx.next_avail_desc`.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+    ///   pseudo-header checksum (the NIC adds the segment checksum on top), or `None`.
+    fn igc_tx_ctx_setup(
+        &self,
+        tx: &mut Tx,
+        ether_frame: &net_device::EtherFrameRef,
+        head: usize,
+    ) -> Result<(usize, u32, Option<L4CksumSeed>), IgcDriverErr> {
+        let base_olinfo = (ether_frame.data.len() as u32) << IGC_ADVTXD_PAYLEN_SHIFT;
+
+        let ext = match extract_headers(ether_frame.data) {
```
atsushi421: `extract_headers()` is parsed on every send, walking Ethernet/VLAN/IP/TCP/UDP with several unsafe reads, even when `csum_flags` is empty and no offload is requested. On the fast path this is the dominant per-packet cost added by this PR.

Early-return at the top: `if ether_frame.csum_flags.is_empty() { return Ok((0, base_olinfo, None)); }`.

awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -933,27 +956,161 @@ impl IgcInner {
...
+        // The per-packet pseudo-header seed is still required, so return `cksum_seed`.
+        if tx.active_checksum_context == new_ctx {
+            return Ok((0, olinfo_status, cksum_seed));
+        }
+
+        // Write a new context descriptor at head.
+        type_tucmd_mlhl |= IGC_ADVTXD_DTYP_CTXT | IGC_TXD_CMD_DEXT;
+        let mss_l4len_idx = l4len << IGC_ADVTXD_L4LEN_SHIFT;
+
+        let desc = &mut tx.tx_desc_ring.as_mut()[head];
```
atsushi421, line 1132: `igc_tx_ctx_setup` writes the context descriptor and updates `tx.active_checksum_context` before `igc_send` confirms the ring has room (the in-code comment at L1142-L1143 acknowledges this), which is why the caller has to snapshot and restore `saved_ctx` on the bail-out path. Functionally it is fine — `next_avail_desc` does not move so the slot is not booked, the next `igc_send` overwrites the 16-byte descriptor in full, and the stale `mss_l4len_idx` overlapping `wb.status` keeps the DD bit at 0 so `igc_txeof` is not misled — but the side-effect boundary is fragile and the rollback is easy to forget when extending the function later.

Two cleaner options:
- Make `igc_tx_ctx_setup` pure: take `active_checksum_context` by value and return `(new_ctx, Option<ctx_payload>, olinfo_status, cksum_seed)`. `igc_send` commits both the descriptor write and the state update only after the room check.
- Or do the conservative early check that the OpenBSD igc driver uses: at the top of `igc_send`, bail out if `igc_desc_unused() < 2` (the maximum a single packet can consume). This removes the snapshot/restore dance with minimal churn.

awkernel_drivers/src/pcie/intel/igc.rs

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed. Thank you.

awkernel_drivers/src/pcie/intel/igc.rs

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed.
Check `igc_desc_unused() < 2`.
Thank you.

awkernel_drivers/src/pcie/intel/igc.rs
ytakano: Fixed.
Gate the flags on IPv4 in `tx_packet_header_flags()`.
Thank you.

atsushi421 (approved):

PR #695, opened 2026-06-12
fix(igc): do not use `num_cpu()` when slecting the number of queues

ytakano:
## Description

`num_cpu()` is not available when initializing `igc`.
So, this PR removes `num_cpu()` usage in `igc_select_num_queues()`.

## Related links

#335 

## How was this PR tested?

Tested on a physical machine.

### DIFF
--- a/awkernel_drivers/src/pcie/intel/igc.rs
+++ b/awkernel_drivers/src/pcie/intel/igc.rs
@@ -1494,17 +1494,15 @@ fn igc_is_valid_ether_addr(addr: &[u8; 6]) -> bool {
 
 /// Select the number of Rx/Tx queue pairs to enable.
 ///
-/// The result is constrained by three factors:
+/// The result is constrained by two factors:
 /// 1. **MSI-X vectors**: one vector per queue plus one for link events.
-/// 2. **CPU count**: no benefit in having more queues than CPUs.
-/// 3. **Hardware cap**: IGC/I225 supports at most 4 RSS queues.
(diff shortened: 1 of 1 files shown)


awkernel_drivers/src/pcie/intel/igc.rs
```diff
@@ -1494,17 +1494,15 @@ fn igc_is_valid_ether_addr(addr: &[u8; 6]) -> bool {
 
 /// Select the number of Rx/Tx queue pairs to enable.
 ///
-/// The result is constrained by three factors:
+/// The result is constrained by two factors:
 /// 1. **MSI-X vectors**: one vector per queue plus one for link events.
```
atsushi421, line 1498: The doc comment correctly drops the CPU-count factor, but the rationale (num_cpu() is not yet available during igc initialization) only lives in the PR description. A future maintainer could reasonably re-introduce a CPU-based cap and hit the same init-order bug. Consider adding a short note in the doc comment explaining why the CPU count is intentionally not used as a bound here.

ytakano: Add a doc comment as follows.

```rust
/// Note: for future extensibility
///
/// `num_cpu()` is not available when this function is called during device initialization.
/// If the number of CPU cores is needed to determine the number of queues, the driver should
/// find the number of CPU cores through other means (e.g., ACPI tables).
```

atsushi421 (approved):

PR #696, opened 2026-06-14
feat(alloc): support wait-free allocator

ytakano:
This PR introduces `wf_alloc` as a feature-selectable heap allocator backend for `awkernel_lib`, while keeping TLSF as the fallback path on architectures where `wf_alloc` is not enabled/supported. It also updates heap allocator documentation to match current behavior.

## Description

- Make heap backend selection explicit by compile-time feature:
  - `heap-wf-alloc` selects `wf_alloc`.
  - otherwise, TLSF is used.
- Keep allocator behavior under `Talloc` policy (`primary`/`backup`) unchanged:
  - user-space path still uses primary-first behavior.
  - kernel path uses primary-then-backup behavior.
  - deallocation routing remains range-based.
- Add `wf_alloc` plumbing for supported targets (`x86_64`, `aarch64`):
  - add optional dependency and feature gating in `awkernel_lib`.
  - add compile-time guard that rejects `heap-wf-alloc` on unsupported architectures.
  - implement backend switch via `type Allocator = ...` alias.
- Update architecture heap initialization to pass active CPU count into allocator setup:
  - x86_64 and aarch64 now initialize primary/backup heaps via `init_*_with_num_cpu(...)`.
- Default feature wiring for `kernel` is aligned with this backend selection (`heap-wf-alloc` on x86 by default, TLSF remains for non-CAS2 paths).
- Update `awkernel_lib/src/heap.rs` doc comments:
  - remove stale claim that TLSF is default.
  - document `wf_alloc` initialization flow, CPU-token mapping, interrupt-guarded alloc/dealloc, and fallback to existing OOM path when allocator init fails.
- Initialize `AcpiTable` before heap memory initialization for x86_64.

## Related links

Benchmarking:

- https://ytakano.github.io/wf_alloc/memalloc_report_20260613_135953.html
- wf_alloc is faster than mimalloc and TLSF
- wf_alloc's alloc/free worst case latency is better mimalloc and TLSF

## How was this PR tested?

Tested on Qemu and physical machines


awkernel_lib/src/heap.rs
```diff
@@ -289,73 +390,194 @@ impl Talloc {
...
-    pub unsafe fn init(&self, heap_start: usize, heap_size: usize) {
-        let mut node = MCSNode::new();
-        let mut guard = self.0.lock(&mut node);
-        init_heap(&mut guard, heap_start, heap_size);
+    impl HeapBackend for WfAllocBackend {
+        unsafe fn init(&self, heap_start: usize, heap_size: usize, active_threads: usize) {
+            if self.initialized.load(Ordering::Acquire) {
+                return;
+            }
+
```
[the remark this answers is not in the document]
ytakano: This comment is nonsense.
`init()` is used for the heap memory initialization and it must be called once during kernel booting.

kernel/src/arch/x86_64/kernel_main.rs
```diff
@@ -133,6 +140,21 @@ fn kernel_main(boot_info: &'static mut BootInfo) -> ! {
     wait_forever();
 }
 
+fn detect_num_cpus(boot_info: &BootInfo, offset: u64) -> Option<usize> {
+    let acpi = awkernel_lib::arch::x86_64::acpi::create_acpi(boot_info, offset)?;
+    let madt = acpi.find_table::<acpi::madt::Madt>().ok()?;
+    let num_cpus = madt
```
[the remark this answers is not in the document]
ytakano: Fixed.
Initialize AcpiTable before initializing heap memory allocation.

atsushi421 (commented):
The mdbook chapter at `mdbook/src/internal/memory_allocator.md` still describes the old TLSF-only design (Allocator/BackUpAllocator structs, `init_primary`/`init_backup` only). It should be updated to cover the new `HeapBackend` trait, backend selection by feature, and the `*_with_num_cpu` APIs (and explain why an explicit `num_cpu` is needed for wf_alloc).

---

The new ~200 lines of layout math in `WfAllocBackend::init` (alignment, overflow guards, metadata sizing, `active_threads` validation) have no unit tests. `align_up` alone is a pure helper that is trivially testable. Consider adding `#[cfg(test)] mod tests` that at least covers `align_up` overflow/normal cases and the rejection paths of `init` (too small heap, `active_threads == 0` or `> NUM_MAX_CPU`, misaligned `heap_start`).

---

`is_primary_mem` (around line 384 in `awkernel_lib/src/heap.rs`) reads `primary_start` and `primary_size` with `Relaxed`, and the matching stores in `init_*_with_num_cpu` are also `Relaxed`. Another CPU can observe "new `primary_start`, old `primary_size = 0`" between the two stores; in that window `end == start`, the range check is always false, and primary-owned deallocations are routed to the backup, which would treat them as foreign and corrupt its free list. Use `Release` for the size store (last) and `Acquire` for the load, or document that init must happen-before any allocation.

awkernel_lib/src/heap.rs
```diff
@@ -289,73 +390,194 @@ impl Talloc {
...
+            unsafe { allocator.alloc_with_token(layout, token) }
+        }
+
+        unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+            let Some(allocator) = self.allocator() else {
+                return;
+            };
+            let _interrupt_guard = InterruptGuard::new();
+            let cpu_id = cpu::cpu_id();
+            if cpu_id >= allocator.active_threads() {
```
atsushi421, line 687: When `cpu_id() >= allocator.active_threads()`, this silently returns and the block is permanently leaked, while `alloc` for the same condition only returns null (which the caller can signal). A silent dealloc leak in a kernel allocator is hard to detect at runtime. At minimum this should panic with a message that includes `cpu_id` and `active_threads` so a misconfiguration is caught at the offending dealloc instead of as an unrelated symptom later.

awkernel_lib/src/heap.rs
```diff
@@ -289,73 +390,194 @@ impl Talloc {
...
+        }
     }
 
-    pub unsafe fn init(&self, heap_start: usize, heap_size: usize) {
-        let mut node = MCSNode::new();
-        let mut guard = self.0.lock(&mut node);
-        init_heap(&mut guard, heap_start, heap_size);
+    impl HeapBackend for WfAllocBackend {
+        unsafe fn init(&self, heap_start: usize, heap_size: usize, active_threads: usize) {
+            if self.initialized.load(Ordering::Acquire) {
```
atsushi421, line 606: `init` is `unsafe fn(&self, ...)` and the type implements `Sync`, so concurrent callers are not prevented at the type level. This fast-path check is not mutual exclusion: two CPUs can both observe `initialized = false`, both write into the `UnsafeCell`, and both call `allocator.init(...)`, which is UB. Either claim the slot with a `compare_exchange` (false to an intermediate "initializing" state) or document that callers must serialize and add a `debug_assert!` against re-entry.

awkernel_lib/src/heap.rs
```diff
@@ -289,73 +390,194 @@ impl Talloc {
...
+            }
+        }
     }
 
-    pub unsafe fn init(&self, heap_start: usize, heap_size: usize) {
-        let mut node = MCSNode::new();
-        let mut guard = self.0.lock(&mut node);
-        init_heap(&mut guard, heap_start, heap_size);
+    impl HeapBackend for WfAllocBackend {
+        unsafe fn init(&self, heap_start: usize, heap_size: usize, active_threads: usize) {
```
atsushi421: This function has roughly ten `return;` paths (alignment, metadata size, `active_threads` bounds, integer overflow, `from_metadata_region` failure, span size) and none of them log or signal anything. The caller cannot distinguish a successful init from a half-init. By the time the first allocation fails, the boot log already says the heap is up. Consider returning `Result<(), &'static str>` from `HeapBackend::init` so each failure path carries a reason and the wrapper can `panic!("primary heap init failed: {msg}")`. At minimum emit `unsafe_puts("wf_alloc init failed: <reason>\r\n")` on each failure branch.

awkernel_lib/src/heap.rs
```diff
@@ -78,21 +118,58 @@ pub unsafe fn init_primary(primary_start: usize, primary_size: usize) {
...
+/// concurrently use the heap.
+pub unsafe fn init_backup_with_num_cpu(backup_start: usize, backup_size: usize, num_cpu: usize) {
+    TALLOC.init_backup_with_num_cpu(backup_start, backup_size, num_cpu);
+}
+
+#[cfg(all(
+    feature = "heap-wf-alloc",
+    not(any(target_arch = "x86_64", target_arch = "aarch64"))
+))]
+compile_error!("heap-wf-alloc currently requires a wf_alloc CAS2 backend for this architecture");
```
atsushi421, line 181: The existing `compile_error!` only fires for `heap-wf-alloc` on unsupported architectures. If both `heap-tlsf` and `heap-wf-alloc` are enabled at the same time (e.g., by an unrelated downstream crate), the cfg gating silently selects `WfAllocBackend` and `heap-tlsf` is ignored. Consider adding `#[cfg(all(feature = "heap-tlsf", feature = "heap-wf-alloc"))] compile_error!("`heap-tlsf` and `heap-wf-alloc` are mutually exclusive; enable only one");` to make feature-flag mistakes loud.

awkernel_lib/src/heap.rs
```diff
@@ -9,6 +22,20 @@
...
+//! `wf_alloc` initializes `WfSpanAllocator` from the raw heap region:
+//! 1. Align `heap_start` by `metadata_region_align()`.
+//! 2. Reserve `metadata_region_size(active_threads)` bytes for metadata.
+//! 3. Align the remaining start by `wf_alloc::SPAN_ALIGN`.
+//! 4. Construct the allocator with `from_metadata_region(...)` and initialize it with `init(...)`.
+//!
+//! CPU IDs are mapped directly to `wf_alloc` tokens, so allocation/deallocation
+//! is done with `InterruptGuard` and `alloc_with_token` / `dealloc_with_token`.
+//! If `wf_alloc` cannot be initialized, its allocator stays uninitialized and
+//! all alloc calls return `null`, so `Talloc` falls back to its existing OOM
```
atsushi421: This sentence misrepresents the actual behavior. `Talloc::alloc` calls `panic!()` on a null return in the `is_primary()` branch and falls through to `unsafe_puts` + `delay::wait_forever()` in the non-primary branch when backup also returns null. There is no graceful OOM handler path. Reword to make clear that a `wf_alloc` init failure means the kernel does not boot (or every userland allocation panics), so readers do not assume an alloc_error_handler-style fallback.

awkernel_lib/src/heap.rs
```diff
@@ -78,21 +118,58 @@ pub unsafe fn init_primary(primary_start: usize, primary_size: usize) {
     TALLOC.init_primary(primary_start, primary_size);
```
atsushi421, line 139: This wrapper now delegates to `init_primary_with_num_cpu(..., cpu::num_cpu())`, but the Safety doc above does not mention the dependency on `cpu::num_cpu()`. On x86_64 the heap is initialized before `set_num_cpu()` is called, so `cpu::num_cpu()` returns 0 and `WfAllocBackend::init` silently bails out at the `active_threads == 0` check. Either document this hazard, mark the wrapper `#[deprecated]`, or remove it. The same applies to `init_backup` below.

kernel/src/arch/x86_64/kernel_main.rs
```diff
@@ -81,10 +81,10 @@ const MPBOOT_REGION_END: u64 = 1024 * 1024;
...
 /// 2. Initialize a serial port.
 /// 3. Initialize the virtual memory.
-/// 4. Initialize the backup heap memory allocator.
-/// 5. Enable logger.
-/// 6. Get offset address to physical memory.
-/// 7. Initialize ACPI.
+/// 4. Initialize ACPI.
+/// 5. Initialize the backup heap memory allocator.
+/// 6. Enable logger.
+/// 7. Get offset address to physical memory.
```
atsushi421: The step list now places ACPI at 4 and backup heap at 5, but `physical_memory_offset` is read twice: once in `kernel_main` (line 119, unlabeled) to call `create_acpi`, and again in `kernel_main2` as step 7. The duplicate read and the unlabeled first read make this doc list misleading. Either fold steps 4-7 so the offset is fetched once and reuse it, or add an inline comment noting the early read.

kernel/src/arch/x86_64/kernel_main.rs
```diff
@@ -116,9 +116,26 @@ fn kernel_main(boot_info: &'static mut BootInfo) -> ! {
...
+    // 4. Initialize ACPI.
+    let acpi = if let Some(acpi) = awkernel_lib::arch::x86_64::acpi::create_acpi(boot_info, offset)
+    {
+        acpi
+    } else {
+        unsafe { unsafe_puts("Failed to initialize ACPI.\r\n") };
+        wait_forever();
+    };
+
+    let early_num_cpu = detect_num_cpus(&acpi).unwrap_or(1);
```
atsushi421: On a system without a MADT (or with no enabled `LocalApic`/`LocalX2Apic`), this silently degrades to `active_threads = 1`. When non-primary CPUs later use the backup heap, they all hit `cpu_id >= active_threads` and either fail allocation or leak deallocations, with the visible symptom being unrelated to the missing MADT. Treat this as a fatal boot error (or at minimum a `log::warn!`) so the actual ACPI problem is the diagnostic.

kernel/src/arch/x86_64/kernel_main.rs
```diff
@@ -127,20 +144,36 @@ fn kernel_main(boot_info: &'static mut BootInfo) -> ! {
...
             backup_region,
             backup_next_frame,
+            acpi,
         )
     });
 
     wait_forever();
 }
 
+fn detect_num_cpus(acpi: &AcpiTables<AcpiMapper>) -> Option<usize> {
```
atsushi421: `detect_num_cpus` filters MADT entries only by the enabled bit, but `non_primary_cpus` later applies extra filters (`ProcessorState::WaitingForSipi`, xAPIC `< 255`). The two counts can legitimately diverge, leaving the backup heap sized for one population and the primary heap for another. Either align this filter with the one used to build `non_primary_cpus`, or defer all heap init until `non_primary_cpus` is known and use that single count everywhere.

kernel/src/arch/x86_64/kernel_main.rs
```diff
@@ -298,7 +322,11 @@ fn kernel_main2(
     }
 
     // 15. Initialize the primary heap memory allocator.
-    init_primary_heap(&mut page_table, &mut page_allocators);
+    init_primary_heap(
+        &mut page_table,
+        &mut page_allocators,
+        non_primary_cpus.len() + 1,
```
atsushi421, line 365: `init_backup_heap` is called with `early_num_cpu` (MADT enabled count) and `init_primary_heap` is called here with `non_primary_cpus.len() + 1` (after `ProcessorState::WaitingForSipi` and xAPIC filters). When these counts differ, primary and backup are sized for different populations and high cpu_ids will fail or leak on whichever allocator was sized smaller. Pass the same value to both initializers.

awkernel_lib/src/heap.rs
ytakano: Thanks. Reworded the module doc so it matches the actual behavior: on a `wf_alloc` init failure the allocator stays uninitialized and every allocation returns `null`, and `Talloc` then either `panic!`s (primary path) or prints via `unsafe_puts` and halts in `delay::wait_forever()` (non-primary path). The doc now states explicitly that there is no `alloc_error_handler`-style fallback, i.e. the kernel does not boot / every userland allocation panics.

awkernel_lib/src/heap.rs
ytakano: Thanks. Documented the hazard in the `# Safety` sections of both `init_primary` and `init_backup`: these wrappers read `cpu::num_cpu()` internally, so they must only be called after the active CPU count is established (after `set_num_cpu()`). The doc now notes that if `cpu::num_cpu()` returns `0` (e.g. on x86_64, where the heap is initialized before `set_num_cpu()`), `WfAllocBackend::init` bails out at the `active_threads == 0` check, and that such callers must use `init_primary_with_num_cpu` / `init_backup_with_num_cpu` with an explicit count instead.

awkernel_lib/src/heap.rs
ytakano: Agreed that a silent dealloc leak is the worst failure mode here, so this now fails loudly. I did not use `panic!` though: the kernel panic handler (`kernel/src/nostd.rs`) allocates (`log::error!` and `unwinding::panic::begin_panic(Box::new(()))`), and the CPU hitting this guard is by definition out-of-range, so that allocation would re-enter this same `cpu_id >= active_threads` guard on the same CPU — causing a double panic / recursion, or halting with the misleading "failed to allocate heap memory" message and losing the diagnostic.

Instead I mirror the existing non-primary OOM path, which deliberately avoids allocation: the dealloc guard now prints `cpu_id` and `active_threads` via the allocation-free `unsafe_puts` / `unsafe_print_hex_u64` and halts in `delay::wait_forever()`. The misconfiguration is therefore caught at the offending dealloc with both values in the message. Fixed in e31a2336.

awkernel_lib/src/heap.rs
ytakano: You're right that the `initialized` fast-path is not mutual exclusion. I took the "document that callers must serialize" option: `HeapBackend::init` now carries a `# Safety` contract stating that `init` must be called exactly once, on a single thread, and before the allocator becomes reachable by any other CPU, and it spells out that because the backend is `&self`/`Sync` the fast-path is not a lock, so concurrent `init` is UB. The kernel satisfies this by initializing the heap during single-threaded boot, before the APs are started (as noted earlier, `init` is called once during boot).

I left out the `debug_assert!(!initialized)` companion on purpose: the top-of-`init` guard intentionally no-ops on re-entry, so a debug-assert against re-entry would contradict that tolerant design. Fixed in 919688b5.

awkernel_lib/src/heap.rs
ytakano: Done, taking the `Result` route you suggested. `HeapBackend::init` now returns `Result<(), &'static str>`, and every failure path in `WfAllocBackend::init` (alignment/overflow guards, metadata sizing, `active_threads` bounds, `from_metadata_region`) returns a descriptive reason instead of a silent `return`. `Talloc::init_{primary,backup}_with_num_cpu` surface the reason on failure, so a half-init is reported at init time rather than as an opaque allocation failure later.

One deviation: instead of `panic!("... {msg}")`, the wrapper prints `unsafe_puts("primary/backup heap init failed: <reason>\r\n")` and halts in `delay::wait_forever()`. The panic handler (`kernel/src/nostd.rs`) allocates (`log::error!`, `unwinding::panic::begin_panic(Box::new(()))`), which cannot work while the heap is the very thing that just failed to initialize — so a `panic!` here would likely halt with the misleading "failed to allocate heap memory" message and lose the reason. `unsafe_puts` is allocation-free, so the actual reason is printed reliably. Fixed in f906a4ee.

awkernel_lib/src/heap.rs
ytakano: Done. Added `#[cfg(all(feature = "heap-tlsf", feature = "heap-wf-alloc"))] compile_error!("`heap-tlsf` and `heap-wf-alloc` are mutually exclusive; enable only one");` next to the existing arch guard. Verified that a normal x86_64 build still compiles and that enabling both features now fails the build with that message. Fixed in 1cfdbce7.

kernel/src/arch/x86_64/kernel_main.rs
ytakano: Folded it. `physical_memory_offset` is now read exactly once in `kernel_main` — labeled step 4 ("Get offset address to physical memory"), placed before ACPI since `create_acpi` needs it — and passed into `kernel_main2` as a parameter, so the duplicate read (old step 7) is gone. The step list is renumbered accordingly (offset=4, ACPI=5, backup heap=6, logger=7; step 8 onward unchanged). Fixed in d6976b16.

kernel/src/arch/x86_64/kernel_main.rs
ytakano: Addressed together with the two related comments below, in 4e9a3dc7. For this one specifically: the no-MADT case is now fatal instead of `unwrap_or(1)` — `count_usable_cpus` (renamed from `detect_num_cpus`) returns `None` when the MADT is missing, and `kernel_main` then prints "Failed to detect CPUs from ACPI (MADT/processor info missing)." and halts, so the real ACPI fault is the diagnostic instead of a later allocation failure on an unsized heap.

kernel/src/arch/x86_64/kernel_main.rs
ytakano: Fixed in 4e9a3dc7 (together with the related comments above/below). Rather than aligning the exact filter, I documented and enforced *why* the divergence is safe: the early count must stay allocation-free because it runs before any heap is initialized — going through `acpi.platform_info()` (as `non_primary_cpus` does) allocates a `Vec` and OOMs there (I hit exactly this while testing). So `count_usable_cpus` keeps the non-allocating `madt.entries()` enabled-bit count, which is a deliberate **upper bound**: `non_primary_cpus` only ever applies *additional* restrictive filters (`WaitingForSipi`, xAPIC `< 255`), so the enabled count is always `>= non_primary_cpus.len() + 1`. The backup heap is sized with this upper bound, so it is never smaller than the primary heap, and `kernel_main2` re-checks the invariant once the exact set is known (see the reply on the next comment).

kernel/src/arch/x86_64/kernel_main.rs
ytakano: Fixed in 4e9a3dc7. Instead of forcing the exact same value into both initializers (impossible at the early call site: `non_primary_cpus` needs both the heap and `type_apic`, neither of which exists yet), the backup heap is sized with `count_usable_cpus` — a provable **upper bound** of `non_primary_cpus.len() + 1` — while the primary heap keeps the exact `non_primary_cpus.len() + 1`. Because the backup is an upper bound, it is never sized smaller than the primary, so no cpu_id exceeds either allocator's `active_threads`. To guard against the upper-bound assumption ever being violated, `kernel_main2` now checks `non_primary_cpus.len() + 1 <= count_usable_cpus(&acpi)` right after the set is built and halts loudly with a log if not, instead of letting a high cpu_id fail allocation later. Verified end-to-end: boots to the shell under QEMU with `-smp 16` (15 APs + primary = 16, matching the backup heap count), no inconsistency or OOM.

ytakano:
> The new ~200 lines of layout math in `WfAllocBackend::init` (...) have no unit tests. `align_up` alone is a pure helper that is trivially testable. Consider adding `#[cfg(test)] mod tests` that at least covers `align_up` overflow/normal cases and the rejection paths of `init` ...

Added in da52eabd. While doing this I hit a structural snag worth noting: the `heap` module is `#[cfg(not(feature = "std"))]` (it installs the `#[global_allocator]`), and the host test build runs with `--features std`, so a `#[cfg(test)] mod tests` placed inside `heap` / `wf_alloc_backend` is never compiled by `make test`.

So I extracted `align_up` into a new always-declared `heap_util` module (its body gated to the wf backend or `cfg(test)`), which the std test harness does compile and run — mirroring how `local_heap`'s tests run on the host. The tests cover `align_up`'s already-aligned, round-up, alignment-of-one, large power-of-two, and overflow (`None`) cases (`make test`: 4 new tests, all passing).

I did **not** unit-test the `init` rejection paths: they depend on `wf_alloc` types (`WfAlloc::metadata_region_align/size`, `from_metadata_region`), so testing them on the host would require enabling `heap-wf-alloc` for the whole test binary, which also swaps in an uninitialized `WfAllocBackend` as the global allocator and breaks every other test. If you'd like that coverage, I think the right move is a dedicated wf-enabled test target rather than changing the default one — happy to do that as a follow-up if you want.

ytakano:
> `is_primary_mem` (around line 384 in `awkernel_lib/src/heap.rs`) reads `primary_start` and `primary_size` with `Relaxed` (...) Another CPU can observe "new `primary_start`, old `primary_size = 0`" between the two stores (...) Use `Release` for the size store (last) and `Acquire` for the load, or document that init must happen-before any allocation.

Took the "document that init must happen-before any allocation" option, in 471c5cd3 — and it turns out that's the technically accurate fix here rather than just the convenient one.

The window you describe ("new `primary_start`, old `primary_size == 0`") only exists while the BSP runs heap init single-threaded; the APs never read `primary_start` / `primary_size` until after init has fully completed, because the boot sequence puts a happens-before edge between init and any AP allocation:
- **x86_64**: the BSP does all heap init while the APs are parked, then publishes `BSP_READY` with a release store; each AP observes it and runs an `fence(Acquire)` before its first allocation (`use_primary_then_backup` etc.).
- **aarch64**: the BSP publishes `PRIMARY_INITIALIZED` with a `SeqCst` store after heap init, and each AP spins on a `SeqCst` load of it before proceeding.

So no other CPU ever observes the intermediate state, and a primary-owned deallocation is never misrouted to the backup. I also realized that bumping the two stores/loads to `Release`/`Acquire` wouldn't actually close the hole on its own: `primary_start` and `primary_size` are two separate atomics and can't be read as one atomic pair regardless of ordering — it's the init-before-sharing contract, not the per-atomic ordering, that provides safety. I documented this on `init_primary_with_num_cpu` / `init_backup_with_num_cpu` and `is_primary_mem`, tying it to the single-init `# Safety` contract on `HeapBackend::init`.

Happy to additionally bump these to `Release`/`Acquire` as defense-in-depth if you'd prefer the local atomics to be self-describing, but it isn't required for correctness given the boot handshake.

ytakano:
> The mdbook chapter at `mdbook/src/internal/memory_allocator.md` still describes the old TLSF-only design (Allocator/BackUpAllocator structs, `init_primary`/`init_backup` only). It should be updated to cover the new `HeapBackend` trait, backend selection by feature, and the `*_with_num_cpu` APIs (and explain why an explicit `num_cpu` is needed for wf_alloc).

Updated in 94bc5e2c. The chapter now covers:
- compile-time backend selection (`heap-wf-alloc` → wf_alloc on x86_64/aarch64, otherwise TLSF; mutually exclusive via `compile_error!`);
- the `HeapBackend` trait and the `type Allocator = ...` alias, plus the real `Talloc { primary: Allocator, backup: Allocator, ... }` (replacing the old `Allocator` / `BackUpAllocator` structs);
- both the bare `init_primary` / `init_backup` and the `init_*_with_num_cpu` APIs, with a paragraph on why `wf_alloc` needs the explicit CPU count (it maps each CPU id to a per-CPU token and sizes its metadata from `active_threads`, bailing out on `0`), why the bare variants read `cpu::num_cpu()` and so the boot code uses the explicit-count variants on x86_64, and that TLSF ignores the count;
- the x86_64 / aarch64 boot snippets now passing `num_cpu`.

Also fixed the "For x86_64" copy-paste in the AArch64 section and the `Tallock` typo. `mdbook build` renders cleanly.

This was the last open item from the review — thanks for the thorough pass.

awkernel_lib/src/heap.rs
```diff
@@ -9,6 +22,25 @@
...
+//! CPU IDs are mapped directly to `wf_alloc` tokens, so allocation/deallocation
+//! is done with `InterruptGuard` and `alloc_with_token` / `dealloc_with_token`.
+//!
+//! If `wf_alloc` cannot be initialized, its allocator stays uninitialized and
+//! every allocation returns `null`. There is no graceful, `alloc_error_handler`-style
+//! recovery: when `Talloc::alloc` observes the `null`, the primary path calls
+//! `panic!()`, and the non-primary path prints an error via `unsafe_puts` and then
+//! halts in `delay::wait_forever()`. In other words, a failed `wf_alloc`
+//! initialization means the kernel does not boot (or every userland allocation
+//! panics), rather than falling back to some other OOM behavior.
```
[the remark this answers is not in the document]
ytakano: Thanks — addressed in the latest push. The module docs were rewritten to match the actual behavior: a failed `wf_alloc` initialization no longer reads as "leaves the allocator uninitialized / observed later during allocation." The docs now state that `init_primary_with_num_cpu` / `init_backup_with_num_cpu` report the `Err` immediately (`unsafe_puts` + `delay::wait_forever()`), so the kernel halts at init time and never reaches a later allocation. The separate runtime-OOM path (null → `panic!` / halt after backup is also exhausted) is now documented on its own.

atsushi421 (approved):

Issue #362, opened 2025-03-26
Graceful shutdown


ytakano:
Support graceful shutdown.

PR #698, opened 2026-06-16
feat(async): add task::kill() for force-terminating async tasks

ytakano:
## Description

Adds `task::kill(task_id)` to `awkernel_async_lib`, which marks a task as `Terminated` and removes it from the global TASKS registry. Also fixes a race in `run_main()` where `Poll::Pending` could overwrite a `Terminated` state set by a concurrent `kill()` call.

Adds files for AI agents also.

## Related links

Issue https://github.com/tier4/awkernel/issues/362 and https://github.com/tier4/awkernel/issues/363.

### Qemu Test

Tested on a Qemu VM by using `applications/tests/test_task_kill`.

```text
Welcome to Awkernel!

You can use BLisp language as follows.
https://ytakano.github.io/blisp/

> (factorial 20)
2432902008176640000
> (+ 10 20)
30

Enjoy!

> [         5112 INFO] TASK_KILL_TEST start
[         5134 INFO] TASK_KILL_TEST kill_sleeping_task: PASS (kill returned true)
[         5135 INFO] TASK_KILL_TEST task_removed_from_registry: PASS
[         5135 INFO] TASK_KILL_TEST kill_sleeping_task: PASS (resources freed)
[         5135 INFO] TASK_KILL_TEST kill_unknown_id: PASS
[         5186 INFO] TASK_KILL_TEST kill_idempotent: PASS (second kill returned false)
[         5237 INFO] TASK_KILL_TEST resources_freed_after_kill: PASS
[         5237 INFO] TASK_KILL_TEST kill_preempted_task: step 1 start
[         5239 INFO] TASK_KILL_TEST kill_preempted_task: step 2 spawned id=3
[         5290 INFO] TASK_KILL_TEST kill_preempted_task: step 3 after sleep 50ms
[         5290 INFO] TASK_KILL_TEST kill_preempted_task: step 4 kill=true
[         5591 INFO] TASK_KILL_TEST kill_preempted_task: step 5 after sleep 300ms
[         5591 INFO] TASK_KILL_TEST kill_preempted_task: PASS
[         5592 INFO] TASK_KILL_TEST kill_panicked_task: step 1 start
[         5592 ERROR] kernel/src/nostd.rs:162: panic: panicked at applications/tests/test_task_kill/src/lib.rs:171:13:
intentional panic for kill semantics test
[         5598 INFO] 00:        0x000000005c4b03 - UNKNOWN
[         5600 INFO] 01:        0x000000004313e8 - UNKNOWN
[         5602 INFO] 02:        0x0000000061097b - UNKNOWN
[         5604 INFO] 03:        0x00000000430ed1 - UNKNOWN
[         5607 INFO] 04:        0x0000000059869d - UNKNOWN
[         5608 INFO] 05:        0x0000000059661b - UNKNOWN
[         5610 INFO] 06:        0x000000004261ae - UNKNOWN
[         5610 INFO] 07:        0x0000000042ce1f - UNKNOWN
[         5610 INFO] 08:        0x000000000000fb - UNKNOWN
[         5692 INFO] TASK_KILL_TEST kill_panicked_task: PASS (panicked task removed from registry)
[         5693 INFO] TASK_KILL_TEST kill_panicked_task: PASS (kill returned false)
[         5694 INFO] TASK_KILL_TEST done
```

### SPIN Model Chcker

Update `specification/awkernel_async_lib/src/task/preemptive_spin` to handle `kill()` and tested it works correctly.

### DIFF
--- a/applications/tests/test_task_kill/src/lib.rs
+++ b/applications/tests/test_task_kill/src/lib.rs
@@ -27,6 +27,7 @@ pub async fn run() {
     kill_idempotent().await;
     resources_freed_after_kill().await;
     kill_preempted_task().await;
+    kill_panicked_task().await;
 
     log::info!("TASK_KILL_TEST done");
 }
@@ -145,6 +146,34 @@ async fn kill_preempted_task() {
             "TASK_KILL_TEST kill_preempted_task: FAIL (drop_count={})",
             DROP_COUNT.load(Ordering::Acquire)
         );
+    }
+}
+
+async fn kill_panicked_task() {
+    log::info!("TASK_KILL_TEST kill_panicked_task: step 1 start");
+
+    let id = task::spawn(
+        "panicked-target".into(),
+        async {
+            panic!("intentional panic for kill semantics test");
+        },
+        SchedulerType::PrioritizedFIFO(0),
+    );
+
+    // Let the task start and transition to Panicked after panic handling.
+    sleep(Duration::from_millis(100)).await;
+
+    if task::get_task(id).is_none() {
+        log::info!("TASK_KILL_TEST kill_panicked_task: PASS (panicked task removed from registry)");
+    } else {
+        log::error!("TASK_KILL_TEST kill_panicked_task: FAIL (panicked task still in registry)");
+    }
+
+    // Panicked tasks are terminal; kill() should return false and not mutate state.
+    if !task::kill(id) {
+        log::info!("TASK_KILL_TEST kill_panicked_task: PASS (kill returned false)");
+    } else {
+        log::error!("TASK_KILL_TEST kill_panicked_task: FAIL (kill returned true)");
     }
 }
 
--- a/applications/tests/test_task_kill/src/lib.rs
+++ b/applications/tests/test_task_kill/src/lib.rs
@@ -33,17 +33,20 @@ pub async fn run() {
 }
 
 async fn kill_sleeping_task() {
+    DROP_COUNT.store(0, Ordering::Release);
+
     let id = task::spawn(
         "kill-target".into(),
         async {
+            let _tracker = DropTracker;
             loop {
-                sleep(Duration::from_secs(3600)).await;
+                sleep(Duration::from_millis(40)).await;
             }
         },
         SchedulerType::PrioritizedFIFO(0),
     );
 
-    sleep(Duration::from_millis(50)).await;
+    sleep(Duration::from_millis(20)).await;
 
     if task::kill(id) {
         log::info!("TASK_KILL_TEST kill_sleeping_task: PASS (kill returned true)");
@@ -51,12 +54,19 @@ async fn kill_sleeping_task() {
         log::error!("TASK_KILL_TEST kill_sleeping_task: FAIL (kill returned false)");
     }
 
-    sleep(Duration::from_millis(50)).await;
-
     if task::get_task(id).is_none() {
         log::info!("TASK_KILL_TEST task_removed_from_registry: PASS");
     } else {
         log::error!("TASK_KILL_TEST task_removed_from_registry: FAIL (still in registry)");
+    }
+
+    if wait_for_drop_count(1, 10).await {
+        log::info!("TASK_KILL_TEST kill_sleeping_task: PASS (resources freed)");
+    } else {
+        log::error!(
+            "TASK_KILL_TEST kill_sleeping_task: FAIL (drop_count={})",
+            DROP_COUNT.load(Ordering::Acquire)
+        );
     }
 }
 
@@ -182,8 +192,8 @@ async fn resources_freed_after_kill() {
 
     // The task owns a DropTracker and sleeps 200 ms per iteration.
     // sleep() stores the Waker in a timer closure, which holds the last Arc<Task>
-    // reference after kill() removes the task from TASKS.  When the timer fires
-    // the closure drops the Waker → Arc<Task> → 0 → Future → DropTracker::drop().
+    // reference after kill() removes the task from TASKS.  The dropped Waker should
+    // drop the future and trigger the DropTracker.
     let id = task::spawn(
         "drop-tracker-task".into(),
         async {
@@ -200,10 +210,7 @@ async fn resources_freed_after_kill() {
 
     task::kill(id);
 
-    // Wait for the timer to fire (~150 ms remaining) plus a safety margin.
-    sleep(Duration::from_millis(300)).await;
-
-    if DROP_COUNT.load(Ordering::Acquire) == 1 {
+    if wait_for_drop_count(1, 8).await {
         log::info!("TASK_KILL_TEST resources_freed_after_kill: PASS");
     } else {
         log::error!(
@@ -211,4 +218,15 @@ async fn resources_freed_after_kill() {
             DROP_COUNT.load(Ordering::Acquire)
         );
     }
+}
+
+async fn wait_for_drop_count(expected: usize, retries: usize) -> bool {
+    for _ in 0..retries {
+        if DROP_COUNT.load(Ordering::Acquire) >= expected {
+            return true;
+        }
+        sleep(Duration::from_millis(20)).await;
+    }
+
+    DROP_COUNT.load(Ordering::Acquire) >= expected
 }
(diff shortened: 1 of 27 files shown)


awkernel_async_lib/src/task.rs
```diff
@@ -947,6 +950,52 @@ pub fn wake(task_id: u32) {
...
+/// Force-terminate a task by its ID.
+///
+/// Sets the task state to `Terminated` and removes it from the global task registry.
+/// Any subsequent `wake()` call for this task will be a no-op. If the task was in
+/// `Waiting` state, wakers may still hold `Arc<Task>` references; the `Task` is freed
+/// only after those wakers are dropped — the caller is responsible for deregistering them.
+///
+/// Returns `true` if the task was found and killed, `false` if it was not found or was
+/// already in a terminal state.
+pub fn kill(task_id: u32) -> bool {
```
ytakano, line 1028: **MAJOR — `kill()` defers resource reclamation, possibly forever**

`kill()` removes the task from `TASKS` and sets `Terminated`, but it never drops the task's *future*. The `Arc<Task>` (and therefore the future and all resources it owns) is freed only when the **last** `Arc<Task>` clone is dropped — i.e. when every `Waker` clone held outside `TASKS` is dropped.

For a task parked in `Sleep`, that `Waker` lives in the timer delta-list and is dropped only when the timer fires — so reclamation is deferred for *up to the full remaining sleep duration*. In `kill_sleeping_task` the target sleeps for `3600s`, so after `kill()` its memory stays pinned for ~an hour even though the registry check passes.

Worse: a task parked on a `Waker` that never fires (e.g. a channel `recv` with no remaining sender, or any custom future that stashed its waker elsewhere) is **leaked permanently** — invisible, since it is already out of `TASKS`.

The doc says "the caller is responsible for deregistering them", but no API is exposed to deregister internal wakers, so callers generally cannot. Consider having `kill()` proactively reclaim: `try_lock` the `future` mutex and drop / fuse-terminate the inner future to release resources immediately. Note the in-flight-poll case (future lock held by a polling CPU): if `try_lock` fails, the `Poll::Pending` arm in `run_main` (around `task.rs:884-895`) currently also does not drop the future, so resources still are not freed — that arm should drop / terminate the future when it observes `Terminated`.

awkernel_async_lib/src/sleep_task.rs
```diff
@@ -47,10 +47,21 @@ impl Future for Sleep {
...
+                            let mut guard = state.lock(&mut node);
+                            if let State::Wait = &*guard {
+                                *guard = State::Finished;
+                                true
+                            } else {
+                                false
+                            }
+                        };
+                        if should_wake {
                             waker.wake();
```
ytakano, line 65: **MAJOR — killed-task Drop chain now runs under the global `SLEEPING` lock**

The timer handler is invoked from `SleepingTasks::wake_task()` (`scheduler.rs:327`, `handler()`), which runs **while the global `SLEEPING` mutex is held** (`scheduler.rs:383-384`). With `kill()`, it is now common for `waker.wake()` here to drop the *last* `Arc<Task>` inline (state is `Terminated` → `Task::wake` returns and drops `self`). That triggers the full `Task` → future → nested-future Drop chain **synchronously, under `SLEEPING`**.

Releasing the sleep `state` lock before `wake()` (good fix!) only covers `Sleep::drop`. The broader invariant is now: *arbitrary user `Drop` code runs while `SLEEPING` is held*. Any `Drop` impl in a task's future that schedules a timer / sleeps / otherwise locks `SLEEPING` will self-deadlock — MCS locks are non-reentrant. `Sleep::drop` is safe today (locks only its own `state`), but this is a fragile, easy-to-violate invariant.

Consider deferring the final `Arc<Task>` drop until after `SLEEPING` is released (e.g. pop the fired handlers, drop the `SLEEPING` guard, then invoke them), and/or documenting the invariant.

awkernel_async_lib/src/task.rs
```diff
@@ -947,6 +950,52 @@ pub fn wake(task_id: u32) {
...
+    let task = {
+        let mut node = MCSNode::new();
+        let tasks = TASKS.lock(&mut node);
+        match tasks.id_to_task.get(&task_id) {
+            Some(t) => t.clone(),
+            None => return false,
+        }
+    };
+
+    // Step 2: Under the info lock, transition state to Terminated.
```
ytakano, line 1040: **MINOR — `kill()` is asynchronous w.r.t. an in-flight poll**

`kill()` returns `true` as soon as it sets `Terminated`, but if the task is being polled on another CPU the future keeps running to the end of that poll (cooperative — expected). Worth documenting that `kill()` does not interrupt an in-flight poll, and (per the resource-reclamation comment) that the future is not necessarily dropped even after that poll completes.

applications/tests/test_task_kill/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+pub async fn run() {
+    sleep(Duration::from_secs(1)).await;
+    log::info!("TASK_KILL_TEST start");
+
+    kill_sleeping_task().await;
+    kill_unknown_id();
+    kill_idempotent().await;
+    resources_freed_after_kill().await;
+
+    log::info!("TASK_KILL_TEST done");
```
ytakano, line 32: **MINOR — test gap & timing fragility**

`kill_sleeping_task` spawns a task that sleeps `3600s` and only asserts registry removal — it cannot observe whether the task/future was actually reclaimed (the timer won't fire for an hour), so it gives false confidence that `kill()` "frees" the task. Only `resources_freed_after_kill` (200ms sleep) exercises reclamation, and it depends on the exact deferred-drop timing.

All cases also use fixed wall-clock sleeps (50/300ms), which are timing-fragile under load or a different timer resolution. Consider a shorter sleep + an explicit reclamation assertion for the "sleeping task" case.

awkernel_async_lib/src/task.rs
```diff
@@ -947,6 +950,52 @@ pub fn wake(task_id: u32) {
     gurad.wake(task_id);
 }
 
+/// Force-terminate a task by its ID.
+///
+/// Sets the task state to `Terminated` and removes it from the global task registry.
+/// Any subsequent `wake()` call for this task will be a no-op. If the task was in
+/// `Waiting` state, wakers may still hold `Arc<Task>` references; the `Task` is freed
+/// only after those wakers are dropped — the caller is responsible for deregistering them.
```
ytakano: **NIT — doc comment precision**

"the caller is responsible for deregistering them" — clarify *how* (which API), or soften the wording, since no such API is exposed. Also worth noting: `kill()` returns `false` when the task became terminal between Step 1 and Step 2 (natural completion racing the kill), which is correct behavior but currently undocumented.

ytakano:
## Review summary (concurrency focus)

The `kill()` lock protocol (TASKS → info → TASKS, never nested), the `run_main` `Poll::Pending` guard against resurrecting a killed task, and the sleep `state`-lock-release deadlock fix are all correct and carefully reasoned. I also checked that `NUM_TASK_IN_QUEUE` stays balanced (a killed-but-queued task is decremented when popped in `get_next_task`) and that the registry protocol is safe against id reuse — both look fine.

The two substantive concerns are about lifecycle rather than the state machine:

1. **Resource reclamation** — `kill()` never drops the future, so reclamation is deferred until the last `Waker` drops (up to the full sleep duration, or never for a waker that never fires → permanent leak).
2. **Drop-under-`SLEEPING`** — with `kill()` it is now common for the last `Arc<Task>` to be dropped inside the timer handler while the global `SLEEPING` lock is held, so arbitrary user `Drop` code runs under that lock.

Inline comments have details and suggestions.

awkernel_async_lib/src/task.rs
ytakano: Addressed. kill() now tries to drop the task future when it terminates a non-preempted task, and run_main also drops the future when it observes Terminated/Panicked after an in-flight poll returns Pending. For Preempted, kill_pending defers the termination until the next poll boundary, then removes the task from TASKS and drops the future there. I also updated the kill() docs to describe the in-flight poll and terminal-state behavior.

awkernel_async_lib/src/sleep_task.rs
ytakano: Addressed. I kept the Sleep-local state lock release, and also changed scheduler wakeup handling so expired sleep handlers are taken while SLEEPING is locked but invoked only after the SLEEPING guard is dropped. That avoids running waker.wake() and the possible final Arc<Task>/future Drop chain under the global SLEEPING lock.

PR #701, opened 2026-07-07
feat(sched): add CPU core affinity via ClusteredEDF scheduler

ytakano:
## Description

Generalizes the `PartitionedEDF` scheduler — which pinned each task to a single core — into a **`ClusteredEDF`** scheduler where each task carries a `CpuSet` describing the set of cores it may run on. Pinning to a single core is now the special case of a one-bit `CpuSet`.

Key changes:

- **Affinity-aware run queue**: replaces the per-core `[Mutex<EDFData>; NUM_MAX_CPU]` `BinaryHeap` array with a single affinity-aware priority queue from the [`affinity_btree_queue`](https://github.com/ytakano/affinity_btree_queue) crate, **vendored in-tree as a workspace member** (like `smoltcp`). `get_next` pops the earliest-deadline task runnable on the calling CPU via `pop_for_cpu`. Priority is `(absolute_deadline, wake_time)`, matching the previous EDF ordering, with FIFO tie-breaking on fully-equal keys.
- **`CpuSet`**: `awkernel_lib`'s `CpuSet` is now a type alias of `affinity_btree_queue::CpuMask<CPU_SET_WORDS>` (512 CPUs), so schedulers pass it to the run queue with no conversion. All constructors are `const fn` (chainable builder `CpuSet::empty().with(1).with(2)`), and its `iter()` is O(popcount). The CPU-0-exclusion policy stays in `awkernel_lib` as the const free functions `cpu::masked_workers` / `cpu::all_workers`.
- **API**: `SchedulerType::PartitionedEDF(u64, u16)` → `ClusteredEDF(u64, CpuSet)`; `Task.partitioned_core: Option<u16>` → `Task.cpu_set: Option<CpuSet>`. `spawn` masks out CPU 0 (primary core) and out-of-range bits, and falls back to all worker cores (`1..num_cpu()`) with a warning when the resulting set is empty. This normalization/fallback contract is documented on the `SchedulerType::ClusteredEDF` variant and in the mdbook.
- **O(1) clustered-task bookkeeping**: the per-CPU `NUM_CLUSTERED_TASKS_IN_QUEUE` counter array is replaced by a single global `AtomicU32` plus the run queue's exact affinity mask. `ClusteredTask::new`/`take`/`Drop` now do one atomic RMW instead of one per member CPU (up to 511 for a wide-affinity task). `wake_workers` reads the precise set of CPUs with an eligible queued task via the new `Scheduler::queued_cpu_mask()` trait method (default empty; ClusteredEDF implements it with the queue's O(1) `affinity_mask()`).
- **Extensibility**: a new clustered scheduler only needs to (1) wrap queue entries in `ClusteredTask`, (2) implement `queued_cpu_mask()`, and (3) sit in the clustered prefix of `PRIORITY_LIST`. `SchedulerType::is_clustered()` is the single source of truth for the prefix, and a compile-time assertion rejects any reorder that would break it.
- **Boot invariant**: `cpu::sanity_check()` (run on the primary CPU) panics if `num_cpu() < 2`. Awkernel reserves CPU 0 as the primary core and runs tasks only on the worker cores `1..num_cpu()`, so a single CPU has no worker core; enforcing this once at boot replaces a confusing downstream panic and keeps the clustered fallback's messaging valid.
- **Preemption**: `invoke_preemption` selects the core running the lowest-priority task among the set, and skips preemption entirely (enqueue only) when any core in the set is idle.
- **Micro-opt**: `get_scheduler` / `get_priority` take `&SchedulerType` so the `find_map` in `get_next_task` no longer copies the large (`CpuSet`-embedding) `SchedulerType` per iteration.
- **Renames**: `PartitionedTask` → `ClusteredTask`, `NUM_PARTITIONED_TASKS_IN_QUEUE` → `NUM_CLUSTERED_TASKS_IN_QUEUE`, `get_num_partitioned_schedulers` → `get_num_clustered_schedulers`, test crate `test_partitioned_edf` → `test_clustered_edf`.

## Related links

- Crate (upstream / crates.io source): https://github.com/ytakano/affinity_btree_queue — the library is vendored under `affinity_btree_queue/` in this repo.

## How was this PR tested?

- `cargo check_std`, `cargo check_riscv64` (no_std target), `cargo clippy_std` — all pass with no warnings.
- `RUSTFLAGS="-D warnings" make test` — all suites pass, including:
  - `test_awkernel_async_lib` (42),
  - new `cpu.rs` unit tests for the `masked_workers` / `all_workers` worker-mask policy (word boundaries 64/65/128, CPU 0 exclusion, cross-word `iter` ordering),
  - the vendored `test_affinity_btree_queue` crate tests (61 unit + 17 doc).
- `make x86_64 RELEASE=1` — builds successfully.
- Booted in QEMU (`-smp 4`) with the `test_clustered_edf` feature — all `[OK]`, no panic:
  - **core pinning**: single-core `CpuSet` tasks ran only on their assigned core (cores 1–3).
  - **cluster affinity**: a task with `CpuSet {1, 2}` only ran on cores 1 or 2.
  - **EDF preemption**: `light` (earlier deadline) correctly preempted `heavy` on the same core.
  - **multi-core**: tasks on cores 1 and 2 ran in parallel without interference.
  - **empty-set fallback**: tasks spawned with an empty set and with an out-of-range-only set both fell back to worker cores and ran (never on CPU 0).

## Notes for reviewers

- `SchedulerType` grows from 16 bytes to ~72 bytes (still `Copy`) because it embeds a `CpuSet`. It is passed by reference on the hot `get_next_task` path (`get_scheduler`/`get_priority` take `&SchedulerType`) and only by value on the `spawn` path.
- The `get_next_task(false)` path (RR preemption tick, primary CPU only) can never dequeue a clustered task because `spawn` always removes CPU 0 from every set — same invariant as before, noted in `clustered_edf::get_next`.
- **Locking**: the single queue `Mutex` is the innermost lock everywhere (nothing else is acquired while holding it). It is taken under `GLOBAL_WAKE_GET_MUTEX` in `wake_task`/`get_next_task`, and `wake_workers` now also takes it directly (without `GLOBAL_WAKE_GET_MUTEX`) to read `queued_cpu_mask()`; since it is innermost, no lock cycle is introduced. `invoke_preemption` still runs before the queue lock is taken.
- The queue is lazily initialized with `num_cpu()` on first use (`AffinityBTreeQueue::new` is not `const`), following the existing pattern in `prioritized_rr.rs`. `set_num_cpu` runs before the first `spawn`.
- **Known bounded latency (future work)**: when preemption succeeds, the task is pinned to the victim CPU's preemption-pending heap and is invisible to the other CPUs in its set until the IPI is handled; if another core in the set goes idle meanwhile it cannot pick the task up. This is bounded (IPI delivery + the victim's interrupt-disabled sections) and never a lost wakeup — documented in `invoke_preemption`. An enqueue-instead-of-pin strategy would close it but touches the pending-heap machinery shared by all schedulers, so it is deferred.
- `mdbook/src/internal/scheduler.md` is updated to describe ClusteredEDF, the affinity-aware B-tree queue, the global-counter + `queued_cpu_mask` bookkeeping, and the spawn-time normalization/fallback.


docs/print.html
```diff
@@ -923,6 +923,108 @@ <h2 id="x86_64-6"><a class="header" href="#x86_64-6">x86_64</a></h2>
 <h2 id="aarch64-6"><a class="header" href="#aarch64-6">AArch64</a></h2>
 <p>For AArch64, the <code>AArch64</code> structure implements the <code>Mapper</code> trait in <a href="https://github.com/tier4/awkernel/blob/main/awkernel_lib/src/arch/aarch64/paging.rs">awkernel_lib/src/arch/aarch64/paging.rs</a>.
 To handle page tables, the <code>PageTable</code> structure defined in the <a href="https://github.com/tier4/awkernel/blob/main/awkernel_lib/src/arch/aarch64/page_table.rs"><code>awkernel_lib/src/arch/aarch64/page_table.rs</code></a> is used.</p>
+<h2 id="risc-v-32-bit-rv32"><a class="header" href="#risc-v-32-bit-rv32">RISC-V 32-bit (RV32)</a></h2>
+<p>For RV32, the <code>RV32</code> structure implements the <code>Mapper</code> trait in
+<a href="https://github.com/tier4/awkernel/blob/main/awkernel_lib/src/arch/rv32/paging.rs">awkernel_lib/src/arch/rv32/paging.rs</a>.
+Each operation reads the <strong>currently active</strong> page table from the <code>satp</code> register via
+<code>get_page_table</code> (see <a href="internal/arch/../page_table.html">Page Table</a>), then delegates to the Sv32
```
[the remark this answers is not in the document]
ytakano: `docs/print.html` is automatically generated. So, this review can be ignored.

awkernel_async_lib/src/scheduler.rs
```diff
@@ -177,15 +177,15 @@ pub(crate) fn get_next_task(execution_ensured: bool) -> Option<Arc<Task>> {
...
-    if num_partitioned_tasks > 0 {
-        let task = PRIORITY_LIST[..get_num_partitioned_schedulers()]
+    if num_clustered_tasks > 0 {
+        let task = PRIORITY_LIST[..get_num_clustered_schedulers()]
             .iter()
             .find_map(|&scheduler_type| get_scheduler(scheduler_type).get_next(execution_ensured));
 
-        // Decrement is handled by PartitionedTask::Drop inside get_next().
+        // Decrement is handled by ClusteredTask::Drop inside get_next().
         task
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,105 @@ use core::{
...
+
+/// Number of 64-bit words needed to represent a set of `NUM_MAX_CPU` CPUs.
+pub const CPU_SET_WORDS: usize = NUM_MAX_CPU / 64;
+
 static NUM_CPU: AtomicUsize = AtomicUsize::new(0);
 
+/// A set of CPU cores, represented as a bitmask supporting up to `NUM_MAX_CPU` CPUs.
+///
+/// All constructors and predicates are `const fn` so that a `CpuSet` can be
+/// embedded in const-evaluated statics.
```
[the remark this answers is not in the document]
ytakano: Fixed.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,102 @@ use core::{
...
+
+    /// Return the set of all worker CPUs (`1..num_cpus`).
+    pub const fn all_workers(num_cpus: usize) -> Self {
+        let mut set = Self([u64::MAX; CPU_SET_WORDS]);
+        set = set.masked_workers(num_cpus);
+        set
+    }
+
+    /// Iterate over the CPUs contained in the set, in ascending order.
+    pub fn iter(&self) -> impl Iterator<Item = usize> + '_ {
```
atsushi421: `CpuSet::iter` always evaluates the inner `0..64` filter for every word, so each traversal costs a fixed 512 branch evaluations regardless of how many bits are set or how many words are zero. This runs on the hot path (`ClusteredTask::new`/`take`/`Drop` and `invoke_preemption`), so even a single-core pinned task pays 512 evaluations to yield one CPU index. Consider skipping zero words and using `word.trailing_zeros()` with `word &= word - 1` to make the cost O(popcount).

awkernel_async_lib/src/scheduler.rs
```diff
@@ -222,72 +224,57 @@ pub const fn get_priority(sched_type: SchedulerType) -> u8 {
...
     inner: Option<T>,
-    cpu_id: usize,
+    cpu_set: CpuSet,
 }
 
-impl<T> PartitionedTask<T> {
-    pub(crate) fn new(inner: T, cpu_id: usize) -> Self {
-        crate::task::NUM_PARTITIONED_TASKS_IN_QUEUE[cpu_id].fetch_add(1, Ordering::Relaxed);
+impl<T> ClusteredTask<T> {
+    pub(crate) fn new(inner: T, cpu_set: CpuSet) -> Self {
```
atsushi421: `ClusteredTask::new`/`take`/`Drop` each loop over the whole `cpu_set` doing one atomic RMW per member CPU on a distinct cache line. A wide-affinity task (e.g. `CpuSet::any()`, up to 511 workers) does up to 511 atomics per wake, per dequeue, and per drop, versus O(1) in the old partitioned design. Consider whether a single global clustered-task counter plus the queue's affinity-aware `pop_for_cpu` could replace the per-CPU array, or at least ensure the loop only touches set bits.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,102 @@ use core::{
...
+    }
+
+    /// Return word `idx` of the backing bitmask. Panics if `idx >= CPU_SET_WORDS`.
+    pub const fn word(&self, idx: usize) -> u64 {
+        self.0[idx]
+    }
+
+    /// Return a new set keeping only CPUs in `1..num_cpus`
+    /// (CPU 0 is always excluded).
+    pub const fn masked_workers(mut self, num_cpus: usize) -> Self {
```
atsushi421: The new `CpuSet` bit logic (`masked_workers` per-word masking, CPU 0 exclusion, `all_workers`, cross-word `iter`) has no unit tests, even though `awkernel_lib` already unit-tests a comparable bitmap in `net/port_bitmap.rs`. Consider adding a `#[cfg(test)] mod tests` covering `masked_workers` at and around word boundaries (num_cpus of 64/65/128), CPU 0 always excluded, `all_workers` for several sizes including 1, and `iter` ordering across words.

awkernel_async_lib/src/task.rs
```diff
@@ -285,22 +287,23 @@ impl Tasks {
...
                         log::warn!(
-                            "Partitioned core should be between 1 and {}. Falling back to core 1. Given core: {}",
+                            "The CPU set must contain at least one core between 1 and {}. Falling back to all worker cores. Given set: {:?}",
                             num_cpu() - 1,
-                            core
+                            set
                         );
-                        scheduler_type = SchedulerType::PartitionedEDF(deadline, 1);
-                        Some(1u16)
+                        CpuSet::all_workers(num_cpu())
```
atsushi421: The empty-set fallback here (mask to empty, warn, fall back to all worker cores) is a distinct code path that no test exercises; the test crate only spawns valid single-core and `{1,2}` sets. Consider adding a test that spawns a `ClusteredEDF` task with an invalid set (e.g. `CpuSet::empty()`, or a set whose only cores are `>= num_cpu()`) and verifies it still runs on a worker core.

awkernel_async_lib/src/task.rs
```diff
@@ -285,22 +287,23 @@ impl Tasks {
...
-                    if core == 0 || core >= num_cpu() as u16 {
+                // Validate and normalise the CPU set before creating TaskInfo so that
+                // info.scheduler_type and task.cpu_set stay in sync.
+                let cpu_set = if let SchedulerType::ClusteredEDF(deadline, set) = scheduler_type {
+                    // CPU 0 is the primary core and cannot run clustered tasks.
+                    let masked = set.masked_workers(num_cpu());
+                    let normalized = if masked.is_empty() {
                         log::warn!(
-                            "Partitioned core should be between 1 and {}. Falling back to core 1. Given core: {}",
+                            "The CPU set must contain at least one core between 1 and {}. Falling back to all worker cores. Given set: {:?}",
```
atsushi421, line 299: This fallback makes an invalid or empty affinity set runnable on every worker core, a much larger blast radius than the previous PartitionedEDF fallback to a single core 1. The only trace of this policy is this runtime warning; it is not mentioned in the `cpu_set()` doc or the mdbook scheduler section. Consider documenting the fallback so callers know an out-of-range set is normalized to all workers rather than rejected.

awkernel_async_lib/src/scheduler.rs
```diff
@@ -177,27 +177,29 @@ pub(crate) fn get_next_task(execution_ensured: bool) -> Option<Arc<Task>> {
...
-            .find_map(|&scheduler_type| get_scheduler(scheduler_type).get_next(execution_ensured));
-
-        // Decrement is handled by PartitionedTask::Drop inside get_next().
-        task
-    } else {
-        let task = PRIORITY_LIST
+    if num_clustered_tasks > 0 {
+        let task = PRIORITY_LIST[..get_num_clustered_schedulers()]
             .iter()
             .find_map(|&scheduler_type| get_scheduler(scheduler_type).get_next(execution_ensured));
```
atsushi421: `get_scheduler` (and `get_priority`) take `SchedulerType` by value, and this `find_map` copies the full ~72-byte value (now that `CpuSet` is embedded) out of the reference on every iteration, even though only the discriminant is matched. Consider changing these to take `&SchedulerType` and binding a reference in the closures to avoid the per-iteration payload copy.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,102 @@ use core::{
...
+    }
+
+    /// Create a set containing all worker CPUs (`1..num_cpu()`).
+    pub fn any() -> Self {
+        Self::all_workers(num_cpu())
+    }
+
+    /// Create a set from a bitmask of CPUs 0..64.
+    /// Bit `i` of `bits` corresponds to CPU `i`.
+    pub const fn from_bits(bits: u64) -> Self {
```
atsushi421: `from_bits` only writes `words[0]`, so it cannot express any CPU >= 64 even though `CPU_SET_WORDS` supports up to 512. It is also currently unused. Consider accepting `[u64; CPU_SET_WORDS]`, or documenting explicitly that it is limited to the first 64 CPUs (and removing it / `any()` if they are not part of the intended public API).

awkernel_async_lib/src/scheduler.rs
```diff
@@ -177,27 +177,29 @@ pub(crate) fn get_next_task(execution_ensured: bool) -> Option<Arc<Task>> {
...
 
+        // Decrement is handled by ClusteredTask::Drop inside get_next().
```
atsushi421: This comment says the decrement is handled by `ClusteredTask::Drop`, but the decrement actually happens in the explicit `take()` call inside `get_next` (`Drop` is a no-op once `take()` has run). Consider rewording to attribute the decrement to `take()`.

awkernel_async_lib/src/scheduler/clustered_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+                _ => victim = Some((cpu_id, target_task)),
+            }
+        }
+
+        let Some((victim_cpu, target_task)) = victim else {
+            return false;
+        };
+
+        if task > target_task {
+            push_preemption_pending(victim_cpu, task);
```
nokosaaan, line 197: The logic inside `invoke_preemption` (which examines each CPU in the set, selects the CPU with the lowest-priority task currently running or pending, and sends an interrupt to that CPU) is if an interrupt is sent (i.e., preemption succeeds), the new task is not placed in the affinity queue (a shared location visible to multiple CPUs) but is queued exclusively in the holding area (the `preemption_pending` heap) dedicated to the single CPU selected as the victim.

For example, suppose the `CPU set` for task T is `{1, 2}`. Currently, both Core 1 and Core 2 are executing something, but since the task running on Core 1 has a lower priority, Core 1 is selected as the victim. Task T is placed in the pending heap dedicated to Core 1, and an interrupt (IPI) is sent to Core 1.

Even if, in this scenario, Core 2 finishes its currently running task and becomes idle before the interrupt reaches Core 1, Core 2 is completely unaware of the existence of T (since T is not in the affinity queue and, of course, is not in Core 2’s pending heap either). Core 2 determines that “there is no work for it” and goes to sleep; even though Core 2 could have picked up T immediately, T is forced to wait until Core 1 actually processes the interrupt, creating unnecessary wait time (a latency window).

Bounded, but maybe worth a note or a future enqueue-instead-of-pin strategy.

awkernel_async_lib/src/scheduler/clustered_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+                _ => unreachable!(),
+            }
+        };
+
+        // `Tasks::spawn` normalizes the set to worker cores (1..num_cpu()),
+        // so an invalid set here is an internal invariant violation.
+        let cpu_set = task.cpu_set.expect("Task has no CPU set");
+        if cpu_set.is_empty() || cpu_set.masked_workers(num_cpu()) != cpu_set {
+            panic!("ClusteredEDF: CPU set {cpu_set:?} is out of range");
+        }
```
nokosaaan: invalid CPU sets panic at wake time rather than failing at spawn, and two more panics are reachable through no fault of `wake_task`'s own validation:

`push()` may need to grow the B-tree's backing storage; Rust's allocator has no recoverable-error path for that, so an allocation failure goes straight to `alloc_error_handler` (an abort on most bare-metal targets) while both `GLOBAL_WAKE_GET_MUTEX` and the scheduler's own lock are held. Not a regression — the old `BinaryHeap` had the same property — but worth naming since the move to a single shared queue makes this the only queue left in the hot wake path.

the crate itself panics inside `get_next/pop_for_cpu` if a descent finds the OR-aggregated `subtree_affinity` bit set but no eligible entry beneath it — a deliberate fail-loud choice by the crate author (better than silently losing a task forever), but any future regression in its `split/merge/rotate` bookkeeping would surface as a kernel panic rather than a recoverable error.

I found no way to trigger either today, and the only reachable push error from the kernel's own guards is SequenceOverflow (~`2^64` pushes). But together this hot path now carries three independent panics stacked on it. Is an unconditional crash the intended failure mode across the board, or would spawn-time validation plus a documented contract be preferable?

awkernel_async_lib/src/scheduler/clustered_edf.rs
```diff
@@ -0,0 +1,205 @@
...
+            let mut node_inner = MCSNode::new();
+            let mut data = self.data.lock(&mut node_inner);
+            let queue = data.get_or_insert_with(|| EDFQueue::new(num_cpu()));
+            queue
+                .push(
+                    (absolute_deadline, wake_time),
+                    to_cpu_mask(&cpu_set),
+                    ClusteredTask::new(task.clone(), cpu_set),
+                )
+                .expect("ClusteredEDF: failed to push a task");
```
nokosaaan, line 86: invalid CPU sets panic at wake time rather than failing at spawn, and two more panics are reachable through no fault of `wake_task`'s own validation:

`push()` may need to grow the B-tree's backing storage; Rust's allocator has no recoverable-error path for that, so an allocation failure goes straight to `alloc_error_handler` (an abort on most bare-metal targets) while both `GLOBAL_WAKE_GET_MUTEX` and the scheduler's own lock are held. Not a regression — the old `BinaryHeap` had the same property — but worth naming since the move to a single shared queue makes this the only queue left in the hot wake path.

the crate itself panics inside `get_next/pop_for_cpu` if a descent finds the OR-aggregated `subtree_affinity` bit set but no eligible entry beneath it — a deliberate fail-loud choice by the crate author (better than silently losing a task forever), but any future regression in its `split/merge/rotate` bookkeeping would surface as a kernel panic rather than a recoverable error.

I found no way to trigger either today, and the only reachable push error from the kernel's own guards is SequenceOverflow (~`2^64` pushes). But together this hot path now carries three independent panics stacked on it. Is an unconditional crash the intended failure mode across the board, or would spawn-time validation plus a documented contract be preferable?

awkernel_async_lib/src/task.rs
nokosaaan, line 299: with `num_cpu()==1`, the warning says "Falling back to all worker cores" but `all_workers(1)` is empty and the next wake panics; the message also prints `between 1 and 0`. Returning an error from spawn would be cleaner (pre-existing behavior, but the new message obscures it).

awkernel_async_lib/src/scheduler.rs
```diff
@@ -128,19 +128,19 @@ impl SchedulerType {
...
     SchedulerType::PrioritizedRR(0),
     SchedulerType::Panicked,
 ];
 
-/// Return the number of partitioned schedulers in `PRIORITY_LIST`.
-/// Update this function if you add a new partitioned scheduler to `PRIORITY_LIST`.
-const fn get_num_partitioned_schedulers() -> usize {
+/// Return the number of clustered schedulers in `PRIORITY_LIST`.
+/// Update this function if you add a new clustered scheduler to `PRIORITY_LIST`.
+const fn get_num_clustered_schedulers() -> usize {
```
nokosaaan, line 155: correctness of `NUM_TASK_IN_QUEUE` depends on clustered schedulers forming a strict prefix of `PRIORITY_LIST;` a future reorder would make the full scan decrement a counter never incremented for clustered tasks (`u32 wrap → wake_workers` wakes everyone forever). A `debug_assert!`/comment would cheaply guard this.

awkernel_async_lib/Cargo.toml
```diff
@@ -12,6 +12,7 @@ array-macro = "2.1"
 fixedbitset = "0.5.7"
 async-trait = "0.1"
 async-recursion = "1.1"
+affinity_btree_queue = "0.1"
```
nokosaaan: `"0.1"` floats to any future 0.1.x of a single-author, self-published v0.1 crate that is now the scheduler's core data structure. Suggest `=0.1.0`, vendoring, or moving it into the workspace/org.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,102 @@ use core::{
...
+impl CpuSet {
+    /// Create an empty set.
+    pub const fn empty() -> Self {
+        Self([0; CPU_SET_WORDS])
+    }
+
+    /// Create a set containing all worker CPUs (`1..num_cpu()`).
+    pub fn any() -> Self {
+        Self::all_workers(num_cpu())
+    }
```
nokosaaan: Suppose, in the future, another developer decides, “I want this task to run on any core,” and uses `CpuSet::any()`. Based on the name alone, it might seem to mean “no constraints—leave it up to the system,” but what it actually returns is the set of “all worker cores (all cores except CPU0).” In this particular case, since `any()` and `all_workers(num_cpu())` happen to return the same value, there is no actual harm. However, if a developer mistakenly interprets `any` as “any one of them” and uses it that way, it could lead to misuse—unintentionally expanding the task’s execution scope to all cores.

Please consider the following corrections:
If renaming it, use a name that accurately reflects its behavior (e.g., `all_workers_default()`, which clearly indicates it is a shorthand for `all_workers(num_cpu())`).
Alternatively, simply remove it (since it is unused, and if you want to call it, you can simply write `CpuSet::all_workers(num_cpu());` there is little need to keep a thin wrapper around it).

nokosaaan (changes requested):

awkernel_lib/src/cpu.rs
ytakano: CpuSet is now alias of `AffinityBtreeQueue::CpuMask`.
Fixed as follows. Thank you.

https://github.com/ytakano/affinity_btree_queue/blob/477883973dfe28ed4e4af7c2a9ed6d3a5d037392/src/lib.rs#L309-L321

awkernel_lib/src/cpu.rs
ytakano: Removed `from_bits` and `any`. Thank you.

awkernel_async_lib/src/scheduler.rs
ytakano: Fixed. Thank you.

awkernel_lib/src/cpu.rs
ytakano: Removed `any`. Thank you.

awkernel_async_lib/src/scheduler.rs
ytakano: Added a const assert as follows. Thank you.

```rust
const _: () = {
    let mut i = get_num_clustered_schedulers();
    while i < PRIORITY_LIST.len() {
        assert!(
            !PRIORITY_LIST[i].is_clustered(),
            "clustered schedulers must form a strict prefix of PRIORITY_LIST"
        );
        i += 1;
    }
};
```

awkernel_async_lib/src/scheduler.rs
ytakano: The per-CPU counter array has been replaced by two pieces of state, one per reader:

- `NUM_CLUSTERED_TASKS_IN_QUEUE` is now a single global `AtomicU32`, so `ClusteredTask::new`/`take`/`Drop` perform exactly one atomic RMW regardless of how many CPUs the affinity set contains (previously up to 511). The `cpu_set` field and the per-member loop in `ClusteredTask` are gone entirely.
- The per-CPU information ("which CPUs have an eligible queued task?") is no longer duplicated in counters. It is read directly from the run queue: `affinity_btree_queue` 0.1.1 adds `affinity_mask()`, an O(1) accessor returning the B-tree root's OR-aggregated `subtree_affinity`, which is exactly this set and is already maintained by every push/pop.

awkernel_async_lib/src/scheduler/clustered_edf.rs
ytakano: Correct. This couldn't happen with PartitionedEDF, where the victim was by definition the only core that could run the task; with clusters of two or more cores, pinning to the victim's pending heap forecloses the other cores for the duration of the window.

As you note, it is bounded and never a lost wakeup: the victim's `do_preemption` eventually runs the task, and if the victim's running task finishes first, `run_main` re-wakes the pending tasks — at which point the idle check routes it to the affinity queue where any core in the set can pick it up. The trigger is also narrow: all cores in the set must be busy at wake time *and* one must become idle during the IPI flight.

I've documented the window, its bound, and the recovery paths in a comment in `invoke_preemption`.

I agree enqueue-instead-of-pin is the right long-term fix, but I'd like to treat it as future work in a separate PR: the pending-heap machinery in `preempt.rs` is shared by all schedulers, and the heap currently doubles as duplicate-IPI suppression (via `peek_preemption_pending`), which would need an alternative design.

awkernel_lib/src/cpu.rs
```diff
@@ -5,8 +5,105 @@ use core::{
...
+impl CpuSet {
+    /// Create an empty set.
+    pub const fn empty() -> Self {
+        Self([0; CPU_SET_WORDS])
+    }
+
+    /// Create a set containing all CPUs.
+    pub fn any() -> Self {
+        Self::all_workers(num_cpu())
+    }
```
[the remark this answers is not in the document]
ytakano: Removed.

awkernel_lib/src/cpu.rs
ytakano: Since `CpuSet` is now a type alias of `affinity_btree_queue::CpuMask`, the tests are split across two layers:

- The generic bit primitives (`masked_below` at word boundaries, `iter` ordering across words, the const builders) are unit-tested in the crate itself (`cpu_mask_masked_below_word_boundaries`, `cpu_mask_iter`, `cpu_mask_const_builders`, `cpu_mask_remove` in affinity_btree_queue 0.1.1).
- The awkernel-specific worker-mask policy — which the crate does not know about — now has a `#[cfg(test)] mod tests` in `awkernel_lib/src/cpu.rs` covering `masked_workers`/`all_workers`:
  - `all_workers` for sizes 1 (empty — only CPU 0 exists), 2, 4, and the word boundaries 64/65/128;
  - CPU 0 always excluded, at every size;
  - `iter` ascending across the 63→64 word boundary;
  - `masked_workers` normalization (CPU 0 stripped, out-of-range CPUs trimmed, word boundary at 65) and its idempotence on already-valid sets — the invariant `clustered_edf::wake_task` relies on.

Thank you.

awkernel_async_lib/src/task.rs
ytakano: Documented.

The normalization-and-fallback contract is now spelled out where callers see it: on the `SchedulerType::ClusteredEDF` variant (rustdoc) and in the mdbook scheduler section. Both state that at spawn time CPU 0 and out-of-range bits are removed, and that an empty result falls back to all worker cores (`1..num_cpu()`) with a warning rather than being rejected.

awkernel_async_lib/src/task.rs
ytakano: Added a test.

`test_empty_set_fallback` now spawns a `ClusteredEDF` task for both invalid cases you mention — an empty set and a set whose only core is out of range (`CpuSet::empty().with(num_cpu())`) — and asserts each one still runs, on a worker core and never on CPU 0.

awkernel_async_lib/src/task.rs
ytakano: You're right that the num_cpu()==1 case was the real problem behind the misleading message. Awkernel fundamentally requires num_cpu() >= 2 — CPU 0 is the primary core and tasks run only on the worker cores 1..num_cpu(), so with a single CPU there is no worker core and nothing can be scheduled. That invariant wasn't enforced anywhere, so num_cpu()==1 surfaced as a confusing downstream panic at task wake.

Rather than handle it per-spawn, I've added `cpu::sanity_check()` (called from `sanity::check()` on the primary CPU right after `set_num_cpu`) that panics with a clear message if num_cpu() < 2. This makes num_cpu() >= 2 a single source of truth, so:

- the "between 1 and {num_cpu()-1}" message is now always valid (never "between 1 and 0"), and
- the `all_workers(num_cpu())` fallback is always non-empty.

awkernel_async_lib/src/scheduler.rs
ytakano: Fixed. Thank you.

awkernel_async_lib/src/scheduler/clustered_edf.rs
ytakano: **1. Invalid CPU set (`wake_task`).** This is an internal-invariant assertion, not error handling. `Tasks::spawn` is the single point that validates and normalizes the set — it masks to worker cores and falls back to all workers when empty (now documented on `SchedulerType::ClusteredEDF` and in the mdbook), and `task.cpu_set` never changes afterwards. So reaching `wake_task` with an invalid set means the spawn contract was violated internally. I've changed it to `unreachable!` with a comment naming spawn as the guarantor (2b1efe0a), matching the existing `unreachable!()` in the same function. Failing loud here is deliberate: silently enqueuing an unrunnable set would either hit the crate's `InvalidCpu` push error anyway or produce a task that never runs. So yes — for this one an unconditional crash is intended, and spawn is where the actual validation lives.

ytakano: **2. Allocation failure / push error.** The OOM-aborts-under-lock behavior is a property of the whole kernel's global allocator, not specific to this queue — every `Box`/`Vec`/`Arc` allocation has it, and the old per-core `BinaryHeap` did too, as you note. Of the crate's `PushError` variants, `EmptyAffinity`/`InvalidCpu` are ruled out by the invariant in (1), so the only reachable one from the kernel's own guards is `SequenceOverflow` (~2^64 pushes) — effectively unreachable. I've left the `.expect` as-is; it documents that a push failure is a should-never-happen condition.

**3. Crate-internal fail-loud (`pop_for_cpu`).** That panic is a deliberate choice in the crate: a `subtree_affinity` bookkeeping regression is caught immediately rather than silently losing a task. I agree with that as the intended failure mode; it's the crate's contract, and I'd rather it surface than degrade quietly.

awkernel_async_lib/Cargo.toml
ytakano: I've brought the crate in-tree as a workspace member (the same approach as the vendored `smoltcp`): `awkernel_lib` and `awkernel_async_lib` now depend on it via `path = "../affinity_btree_queue"`, so the exact reviewed source is fixed in the repo and there is no external fetch. Only the library is vendored (Cargo.toml, README, src); its unit tests run through a new `test_affinity_btree_queue` alias wired into `make test`. The upstream repo remains the crates.io source and keeps the dev-only spec/reference material.

awkernel_async_lib/src/scheduler/clustered_edf.rs
ytakano: See above comments.

nokosaaan (approved):

awkernel_async_lib/src/scheduler.rs
```diff
@@ -167,81 +207,101 @@ pub(crate) trait Scheduler {
...
 
+        // The counter is decremented by the ClusteredTask::take() call inside
+        // get_next() (Drop is a no-op once take() has run).
         if task.is_some() {
-            crate::task::NUM_TASK_IN_QUEUE.fetch_sub(1, Ordering::Relaxed);
+            return task;
         }
+    }
+
+    let task = PRIORITY_LIST
```
atsushi421: When the clustered prefix above returns None, this fallback re-runs the clustered schedulers' get_next. The queue state is frozen under GLOBAL_WAKE_GET_MUTEX, so the second call deterministically returns None, at the cost of an extra queue-mutex acquisition and an O(log N) affinity probe on every get_next_task. Iterating `PRIORITY_LIST[get_num_clustered_schedulers()..]` here would skip the already-tried prefix and also structurally rule out a clustered task ever reaching the NUM_TASK_IN_QUEUE decrement below.

mdbook/src/internal/scheduler.md
```diff
@@ -63,22 +63,26 @@ Some schedulers are implemented under the folder [awkernel_async_lib/src/schedul
 
 ```shell
 $ ls awkernel_async_lib/src/scheduler
-> gedf.rs  panicked.rs  partitioned_edf.rs  prioritized_fifo.rs  prioritized_rr.rs
+> clustered_edf.rs  gedf.rs  panicked.rs  prioritized_fifo.rs  prioritized_rr.rs
 ```
 
 A scheduler can be implemented by implementing `Scheduler` Trait.
 Each scheduler must be registered in the following three locations.
 `fn get_next_task()`, `fn get_scheduler(sched_type: SchedulerType)` and `pub enum SchedulerType`.
```
atsushi421: This PR changes the signature to `fn get_scheduler(sched_type: &SchedulerType)`, but this line (and the function table near the top of the page) still shows the old by-value signature. Please update both mentions.

affinity_btree_queue/README.md
```diff
@@ -0,0 +1,159 @@
...
+This project is licensed under either of
+
+- Apache License, Version 2.0
+- MIT license
+
+at your option.
+
+## Test Coverage
+
+[coverage.txt](coverage.txt) contains line-by-line coverage data from `cargo llvm-cov` for the latest test run.
```
atsushi421: This links to coverage.txt, but the file is not committed anywhere in the PR, so the link 404s on GitHub. Consider committing the file, noting it must be generated locally, or dropping the link.

atsushi421 (commented):
The branch bundles the Rust nightly-2026-07-20 toolchain bump (#702) and its fixups (ixgbe.rs, smoltcp tcp.rs, fatfs dir_entry.rs, pvclock.rs, delay.rs, if_net.rs, Dockerfile, CI workflow), which are unrelated to CPU affinity and inflate the review surface. Consider landing #702 on main first and rebasing this PR so those changes drop out of the diff, or stating in the description that the newer nightly is a prerequisite.

---

The docs/ regeneration carries pages unrelated to this PR (RV32/64 mapper, page_table, interrupt_controller, memory_allocator, plus the book.toml and mdbook/src/LICENSE.md changes). Consider regenerating docs from an up-to-date main or limiting the generated-HTML changes to the scheduler pages, and moving the book.toml/LICENSE.md fix to a separate docs PR.

ytakano:
@atsushi421 

> The branch bundles the Rust nightly-2026-07-20 toolchain bump (#702) and its fixups (ixgbe.rs, smoltcp tcp.rs, fatfs dir_entry.rs, pvclock.rs, delay.rs, if_net.rs, Dockerfile, CI workflow), which are unrelated to CPU affinity and inflate the review surface. Consider landing #702 on main first and rebasing this PR so those changes drop out of the diff, or stating in the description that the newer nightly is a prerequisite.

Thank you. Merged main.


> The docs/ regeneration carries pages unrelated to this PR (RV32/64 mapper, page_table, interrupt_controller, memory_allocator, plus the book.toml and mdbook/src/LICENSE.md changes). Consider regenerating docs from an up-to-date main or limiting the generated-HTML changes to the scheduler pages, and moving the book.toml/LICENSE.md fix to a separate docs PR.

I regenerated the docs.
`mdbook` automatically generates mdbook/src/LICENSE.md and RV32/64's pages.
So, can you ignore generated contents for review?

affinity_btree_queue/README.md
ytakano: Drop dead coverage.txt link, note that coverage data is generated locally with cargo llvm-cov.

mdbook/src/internal/scheduler.md
ytakano: Fixed. Thank you,

awkernel_async_lib/src/scheduler.rs
ytakano: Skip the clustered prefix in the fallback loop so it is never re-run (state is frozen under `GLOBAL_WAKE_GET_MUTEX`) and the `NUM_TASK_IN_QUEUE` decrement structurally never applies to a clustered task.

atsushi421 (approved):

